PR #55, opened 2020-03-01
Fix build number from media file on Leap:ARM

andrii-suse:
The problem was caused by two repos matching the same mask.
Thus decided to use explicit name of repo folder in script generation instead of wildcard + arch


andrii-suse:
Closing in favor of https://github.com/andrii-suse/openqa-trigger-from-obs/pull/56 , which fixes the same problem with less changes

PR #78, opened 2020-05-11
Sync JeOS-raspberrypi3 to Leap 15.2 aarch64 images

ggardet:
openQA is ready for this change.


ggardet:
Not sure why CI breaks now with your additional commit.

andrii-suse:
It looks that for some reasons the test was executing longer than expected and status of the job was still `active` instead of expected `finished`. I've restarted the test and not it finished, so PR is fine to go.
```++ echo 'select state from minion_jobs where task='\''obs_rsync_run'\'';'
+ state=' active'
++ echo active
+ test active == finished

Issue #81, opened 2020-05-13
MicroOS 15.2 for aarch64


lnussel:
MicroOS for 15.2 is now also built for aarch64. Could you please include that in the sync scripts?

Also, x86_64 has an rt flavor of the appliance, could you include that too?

andrii-suse:
This was covered with the linked PR

PR #82, opened 2020-05-13
Add 15.2 MicroOS aarch64

andrii-suse:
#81 
Since aarch belongs to different repository inside OBS (images_ports vs images) - the sync must trigger at different point of time than x86_64, thus it must be different project in ObsPlugin folder list and separate xml file.


andrii-suse:
Closing in favor #83

PR #89, opened 2020-07-24
Add config for Jump project

nilxam:
Add config for Jump project. Don't want to use VERSION=15.2 for jump's media since it would matches to Leap 15.2 medias then be selected to `openSUSE Leap 15` group, `openSUSE Jump` group will be used for Jump tests, therefore picks `Jump:15.2` as the version name.


nilxam:
Hi @andrii-suse could you please review this change? I can't test it in real, the re-gened print_* files looks ok to me though.

The difference compared to Leap 15 would be:
* VERSION is Jump:15.* rather than 15.*
* REPO assets name' prefix is `openSUSE-Jump-15.*` rather than `openSUSE-15.*`

andrii-suse:
My idea is that every project/xml file should have corresponding test, so with any refactoring we know all side effects on every project.
But since it is practically the same as Leap with cosmetic difference - maybe such test is not needed, especially if you've verified that source paths in rsync are good.
But I got impression that Jump will have many architectures - or they will be next step after x86_64 works?

nilxam:
> > But I got impression that Jump will have many architectures - or they will be next step after x86_64 works?
> 
> That's what I at least understood from meeting minutes, yes.

Yes, but we disabled non-x86_64 arches for the time being until BS worker's space issue be resolved(we should get new SSDs next week or so).

nilxam:
BTW I don't have permission in this repo, can't merge it.

andrii-suse:
I am not an admin here, so cannot help with permissions. 
I've added tests and have checked generated paths, all look good. I will just merge it and it should be possible to trigger tests once project appears in OpenQA.

andrii-suse:
it has been deployed:
```
$:/opt/openqa-trigger-from-obs> make openSUSE\:Jump\:15.2\:ToTest
mkdir -p openSUSE:Jump:15.2:ToTest && python3 script/scriptgen.py openSUSE:Jump:15.2:ToTest
Generating scripts for openSUSE:Jump:15.2:ToTest
OK
```

PR #91, opened 2020-08-05
Enable aarch64 ppc64le and s390x to Jump project sync

nilxam:
Multiple archs medias but one ftp repo to be used.


nilxam:
should be ready now, let's give it a try?

PR #92, opened 2020-09-03
JeOS images moved from images/ to appliances/

ggardet:


okurz:
Unfortunately tests failed, see

```
# ERROR: lib/test-in-container-systemd.sh at line 74
not ok 1 - IS $ret == 0 # test execution
FAIL opensuse_leap_15.2_staging_a.sh
F-rsync.sh
Makefile:28: recipe for target 'test_docker' failed
```

this does not look like it would be related to your changes but we should not ignore errors in CI. Can you take a look?

ggardet:
> Unfortunately tests failed, see
> 
> ```
> # ERROR: lib/test-in-container-systemd.sh at line 74
> not ok 1 - IS $ret == 0 # test execution
> FAIL opensuse_leap_15.2_staging_a.sh
> F-rsync.sh
> Makefile:28: recipe for target 'test_docker' failed
> ```
> 
> this does not look like it would be related to your changes but we should not ignore errors in CI. Can you take a look?

TBH, I have no idea why it fails. @andrii-suse any idea?

ggardet:
Having a closer look, there is a: `ERROR: 500 - Internal Server Error` upper in the log of the CI. So, the CI seems to be completely broken.

Martchus:
Seems like within these tests an openQA web UI is running an the client script is used to query it:

```
+ openqa-client --host localhost /api/v1/obs_rsync/openSUSE:Leap:15.2:Staging:A/runs put
ERROR: 500 - Internal Server Error
"<!DOCTYPE html>\n<html lang=\"en\">\n  <head>\n
```

Without the web UI log it is hard to tell the reason for the 500 error. Let's see whether restarting helps. If it helps then I suppose your changes are unrelated and tests are just not reliable. (Unfortunately I don't have any further knowledge about these tests.)

ggardet:
Now CI fails at:
```
# ERROR: lib/test-in-container-systemd.sh at line 74
```

Martchus:
I've been restarting the CI checks on master to cross-check whether it is broken as well.

I've also restarted the CI of this PR from the beginning because maybe my last restart didn't work because I only restarted "from failed".

andrii-suse:
Sorry, I've missed this. So integration CI indeed is broken - I am looking at it, but since PR doesn't touch failing tests - I guess we can accept the PR

Martchus:
> but since PR doesn't touch failing tests

And the CI run I've restarted for master failed as well. So yes, the can likely merge it nevertheless if the changes make sense to you.

andrii-suse:
Basically CI met the same problem as production after recent change in openqa client behavior , I this fix will be accepted   https://github.com/os-autoinst/openQA/pull/3355 , otherwise it will require some work .

PR #94, opened 2020-09-10
move TW PowerPC files in a base sub-folder and add microos

michelmno:
To address the question in factory ML https://lists.opensuse.org/opensuse-factory/2020-09/msg00106.html

I made a first step change in  xml/obs/openSUSE:Factory:PowerPC.xml to use a 'base' batch

Note: after discussion with Andrii following changes are harmless:
    * print_openqa.before, normal as added <news> tag.
    * print_openqa.before, '*images/' in place of 'images/' is harmless.

and added microos batch as second step.


andrii-suse:
> * in print_openqa.before, why new trailed line ?

Xml now has <news> tag added, which corresponds to presence of call to factory-package-news.py in print_openqa

> * in print_openqa.before, '*images/local' with a '*' prefix, while expecting 'images/local'

The bash generation is the darkest magic you can imagine - I didn't have any direct requirements, so wanted to just make bash commands work.  (The goal was to move complexity out from runtime and improve testability and be able to preview potential side effects on every project). I suspect that some projects in <batch> needed wildcard, I can try to find exact reason, but I don't think this is right question to ask as long as it doesn't bring problems.
But all magic comes with a price - feel free to ping me in chat if help is needed.

michelmno:
I discussed with Andrii on the two first commits concerning move to base batch.
Then added a third one to add microos batch.

okurz (approved):

michelmno:
good morning @andrii-suse , when could I expect this pr to be merged and used for o3 ?

andrii-suse:
the change has been deployed. It looks I accidentally restarted `base` tests as well. Sorry for that.

PR #95, opened 2020-09-28
Add openSUSE:Jump:15.2:Images

Vogtinator:


t/obs/openSUSE:Jump:15.2:Images:ToTest/print_openqa.before
```diff
@@ -0,0 +1,54 @@
+/usr/share/openqa/script/client isos post --host localhost \
```
Martchus, line 1: `/usr/bin/openqa-cli` should be used instead of the deprecated client. However, I think this is out-of-scope for this PR.

Vogtinator: Yes, absolutely. All I actually did was add the new `.xml` file and let the scripts do the rest.

PR #97, opened 2020-09-29
Add aarch64 media to openSUSE:Jump:15.2:Images

Vogtinator:
Needed some workarounds, see comments in openSUSE:Jump:15:Images.xml.

PR #98, opened 2020-09-29
Add $arch into assets path in print_rsync_iso.sh

andrii-suse:
Fix strange rsync error while copying assets in #97 by explicitly mentioning architecture in path

### DIFF
--- a/script/scriptgen.py
+++ b/script/scriptgen.py
@@ -537,15 +537,20 @@ echo "# Syncing assets"
 declare -A asset_folders''', f)
         for k, v in self.asset_folders.items():
             self.p("asset_folders[{}]='{}'".format(k, v), f)
-        self.p('''while read src; do
+        self.p('''
+for arch in "${archs[@]}"; do
+  while read src; do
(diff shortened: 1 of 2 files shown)


Martchus (approved):

okurz:
@Vogtinator ?

Vogtinator:
This seems to be a difference in glob handling between sh and rsync. rsync appears to do the glob expansion per path element.
This can be worked around by making the last element also a glob, i.e.:

```
            <!-- Matches both -libvirt- and -libvirt_aarch64- -->
            <libvirt folder="*/*vagrant*libvirt*" filemask="libvirt.*box"/>
 <!-- Build broken ATM
            <virtualbox folder="*/*vagrant*virtualbox*" filemask="virtualbox.*box"/>-->
```

I had to use `libvirt.*box` instead of `*libvirt.box` as the latter is not a valid bash regex, so the `asset_tag` matching did not work.

andrii-suse:
closing for now

PR #100, opened 2020-10-19
Factory:ARM: Add libvirt.box asset to RPi3 image - poo#73564

ggardet:
This should fix the missing asset when testing vagrant on RPi4 - https://progress.opensuse.org/issues/73564


andrii-suse:
hm, so far no projects had custom HDD_ and ASSET_ in the same flavor, so it will not work as is. I will make sure that this combination is supported.

ggardet:
Thanks for your fixes!

PR #102, opened 2020-10-29
Handle debuginfo also for ppc64 and armv7hl - poo#65019

ggardet:
Strip ppc64 and armv7 debuginfo repos.
It will fix https://progress.opensuse.org/issues/65019


Martchus (approved):
These `--exclude`s make a list of all common/popular architectures. I'm wondering whether specifying only the ones the are special would make more sense. But that's of course out of the scope of this PR.

andrii-suse:
@Martchus so, that is specific to `rsync` command and how --include and --exclude work together. And the goal of this implementation was to try to repeat rsync.pl behavior as close as possible to limit source of surprises, so I didn't think much about improving that part yet.
So, if somebody spends time and provides alternative command which just uses --include for needed architectures - it should be easy to generate it instead of current monster.

andrii-suse (approved):

ggardet:
It seems this is not enough, see: https://openqa.opensuse.org/admin/obs_rsync/openSUSE:Factory:ARM:ToTest%7Cbase/runs/.run_last/download/rsync_repo.cmd.log

@andrii-suse  any idea what I missed?

PR #103, opened 2020-11-02
Handle debug also for ppc64 and armv7hl - poo#65019

ggardet:
https://github.com/os-autoinst/openqa-trigger-from-obs/pull/102 fixes `debuginfo` only, this PR handles armv7/ppc64 also for `debug`.


andrii-suse (approved):
Yeah #102 adds the flags to only Media3 (source) :( and this is to fix Media2 (debug)

PR #105, opened 2020-11-05
Add option to skip isos post of a specific batch

jlausuch:
This variable is useful when having a xml with different batches
defined and you want to enable/disable some of them for whatever
reason. By default it will behave as before.

### DIFF
--- a/script/scriptgen.py
+++ b/script/scriptgen.py
@@ -111,6 +111,8 @@ class ActionGenerator:
             batch.distri = node.attrib["distri"]
         if node.attrib.get("checksum","") == "0":
             batch.checksum = 0
+        if node.attrib.get("openqa_post","") == "0":
+            batch.openqa_post = 0
         if node.attrib.get("variable",""):
             batch.variable = node.attrib["variable"]
         if node.tag != 'openQA':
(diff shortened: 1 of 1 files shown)


okurz:
hm, thought about including that in a usage and help message or documentation?

jlausuch:
> hm, thought about including that in a usage and help message or documentation?

sure, if the feature is accepted I will add it to the documentation

andrii-suse:
Batches were introduced to manage groups of `isos post`, e.g. with different DISTRI or when artifacts are synced from different paths or when actions are logically independent (e.g. migration tests)
So I don't think about scenario where we need a batch without `isos post`: most often it is possible to have xml without batches at all or reuse objects between batches.
As was discussed in private - we will try to find another solution for original problem. 
I will close the PR for now, but feel free to reopen it if there are any justifications when the feature may be needed.

jlausuch:
> Batches were introduced to manage groups of `isos post`, e.g. with different DISTRI or when artifacts are synced from different paths or when actions are logically independent (e.g. migration tests)
> So I don't think about scenario where we need a batch without `isos post`: most often it is possible to have xml without batches at all or reuse objects between batches.
> As was discussed in private - we will try to find another solution for original problem.
> I will close the PR for now, but feel free to reopen it if there are any justifications when the feature may be needed.

Yes, agreed. Thanks for all the help and support.

PR #108, opened 2020-11-27
[WIP] Rename MicroOS repository extracted from iso

mloviska:
SCC proxy returns repo URL to openqa in specific format for instance: *http://openqa.suse.de/assets/repo/\#\{full_product_name\}-POOL-\#\{arch\}-Build\#\{buildnumber.upcase\}-Media1\* that needs to respected in case of MicroOS as well

```
cat t/ibs/SUSE:SLE-15-SP2:Update:Products:MicroOS:Staging:B/files_iso.lst
SUSE-MicroOS.x86_64-5.0.0-Default-Build14.10.raw.xz
SUSE-MicroOS-5.0-DVD-x86_64-Build40.3-Media.iso
```

This pulled iso should be extracted as an openQA repo in this format:

```
cat t/ibs/SUSE:SLE-15-SP2:Update:Products:MicroOS:Staging:B/print_rsync_iso.after 
rsync --timeout=3600 -tlp4 --specials dist.suse.de::repos/SUSE:/SLE-15-SP2:/Update:/Products:/MicroOS:/Staging:/B/images/*SUSE-MicroOS.x86_64-5.0.0-Default-Build14.10.raw.xz /var/lib/openqa/factory/hdd/SUSE-MicroOS.x86_64-5.0.0-Default-BuildB.14.10.raw.xz
rsync --timeout=3600 -tlp4 --specials dist.suse.de::repos/SUSE:/SLE-15-SP2:/Update:/Products:/MicroOS:/Staging:/B/images/*SUSE-MicroOS.x86_64-5.0.0-Default-Build14.10.raw.xz.sha256 /var/lib/openqa/factory/other/SUSE-MicroOS.x86_64-5.0.0-Default-BuildB.14.10.raw.xz.sha256
rsync --timeout=3600 -tlp4 --specials dist.suse.de::repos/SUSE:/SLE-15-SP2:/Update:/Products:/MicroOS:/Staging:/B/images/iso/*SUSE-MicroOS-5.0-DVD-x86_64-Build40.3-Media.iso /var/lib/openqa/factory/iso/SUSE-5.0-MicroOS-POOL-x86_64-BuildB.40.3-Media1.iso
rsync --timeout=3600 -tlp4 --specials dist.suse.de::repos/SUSE:/SLE-15-SP2:/Update:/Products:/MicroOS:/Staging:/B/images/iso/*SUSE-MicroOS-5.0-DVD-x86_64-Build40.3-Media.iso.sha256 /var/lib/openqa/factory/other/SUSE-5.0-MicroOS-POOL-x86_64-BuildB.40.3-Media1.iso.sha256
[ -d /var/lib/openqa/factory/repo/SUSE-5.0-MicroOS-POOL-x86_64-BuildB.40.3-Media1 ] || {
    mkdir /var/lib/openqa/factory/repo/SUSE-5.0-MicroOS-POOL-x86_64-BuildB.40.3-Media1
    bsdtar xf /var/lib/openqa/factory/iso/SUSE-5.0-MicroOS-POOL-x86_64-BuildB.40.3-Media1.iso -C /var/lib/openqa/factory/repo/SUSE-5.0-MicroOS-POOL-x86_64-BuildB.40.3-Media1
}
```

### DIFF
--- a/script/cfg.py
+++ b/script/cfg.py
@@ -52,6 +52,14 @@ def rsync_commands(checksum):
         echo "rsync --timeout=3600 -tlp4 --specials PRODUCTISOPATH/${iso_folder[$flavor]}*$src.sha256 /var/lib/openqa/factory/other/$dest.sha256"'''
     return res
 
+def rsync_iso_rename_suse_microos(version):
+    if '5.0' in version:
+        return '''dest=${dest/DVD/POOL}
+                  dest=${dest/Media/Media1}
+                  dest=${dest/MicroOS-5.0/5.0-MicroOS}'''
+    else:
+        return ''
+
 rsync_iso = lambda version, archs, staging, checksum : '''
 archs=(ARCHITECTURS)
 
@@ -65,6 +73,7 @@ for flavor in {FLAVORLIST,}; do
         dest=$src
         ''' + rsync_iso_staging(version, staging) + '''
         asset_folder=hdd
+        ''' + rsync_iso_rename_suse_microos(version) + ''' 
         [[ ! $dest =~ \.iso$  ]] || asset_folder=iso
         [[ ! $dest =~ \.appx$  ]] || asset_folder=other
         ''' + rsync_commands(checksum) + '''


mloviska:
My hack is hideous, @andrii-suse, @jlausuch your feedback is very much appreciated if you have time!

mloviska:
Ideally we should check `distri` and not `version` but `distri`'s value is not available in the related lambda function.

jlausuch:
Cool, nice work!
To give a bit more background information, MicroOS repo is organized a bit different than SLE in `.../images/repo/`, for instance we have 

```
  | SUSE-MicroOS-5.0-aarch64-Media1/ |   
  | SUSE-MicroOS-5.0-aarch64-Media2/ |   
  | SUSE-MicroOS-5.0-aarch64-Media3/ |  
  | SUSE-MicroOS-5.0-x86_64-Media1/ | 
  | SUSE-MicroOS-5.0-x86_64-Media2/ |
  | SUSE-MicroOS-5.0-x86_64-Media3/ |
```

instead of 
```
  | SLE-15-SP3-Module-Basesystem-POOL-x86_64-Media1/ |  
  | SLE-15-SP3-Module-CAP-Tools-POOL-x86_64-Media1/ |  
  | SLE-15-SP3-Module-Certifications-POOL-x86_64-Media1/  |  
  | SLE-15-SP3-Module-Containers-POOL-x86_64-Media1/ |
```

script/cfg.py
```diff
@@ -52,6 +52,14 @@ def rsync_commands(checksum):
         echo "rsync --timeout=3600 -tlp4 --specials PRODUCTISOPATH/${iso_folder[$flavor]}*$src.sha256 /var/lib/openqa/factory/other/$dest.sha256"'''
     return res
 
+def rsync_iso_rename_suse_microos(version):
+    if '5.0' in version:
```
jlausuch, line 56: to make this valid in future releases, should we just check for `5.` ?
Actually we should also make sure that we don't hit false positives like `XYZ5.0` cases..
I found this method that might help
```
version.startswith('5.')
```

or with regex
```
import re
if re.match(r'^5.', version):
```

andrii-suse: Is this functionality for sle? Then maybe just define an empty abstract function here and overload it in the plugin, so we have all sle-specific functionality in one place? E.g. constant `openqa_call_legacy_builds` was added with similar purpose, it may be lambda or method as well.

andrii-suse:
> Ideally we should check `distri` and not `version` but `distri`'s value is not available in the related lambda function.

My idea was to pass into such functions only whatever is needed, so you can just try to add parameter `self.ag.distri` here https://github.com/andrii-suse/openqa-trigger-from-obs/blob/master/script/scriptgen.py#L560 and a bit down as well.

But I believe we must do the same rename in `print_openqa` and actually the tests should theoretically catch if iso is named differently in `print_rsync_iso` and `print_openqa`, this is also why I suggest to add related functionality into the gitlab plugin and there here only related stubs are added if needed.

PR #109, opened 2020-11-27
Fix test_repo_consistent

andrii-suse:
The test ensures that repos copied in print_rsync_repo.sh are referenced in isos post.
Proviously the test was skipped because of syntax typo


t/test_repo_consistent.sh
```diff
@@ -26,10 +26,17 @@ for dir in "$@" ; do
 
 	while read -r line; do
 	    if [[ "$line" =~ $regex ]]; then
-	        echo "$known_destination_repos" | grep -q "${BASH_REMATCH[1]}\$" || { >&2 echo "Destination REPO file wasnt found in print_rsync_repo output {${BASH_REMATCH[1]}}   {$known_destination_repos}"; : $((++errs)); continue 2; }
+            # we excuse REPO_0 if print_rsync_iso has bsdtar (which is indication that REPO_0 comes from print_rsync_iso)
```
okurz, line 29: s/excuse/exclude/?

jlausuch:
LGTM besides `exclude` typo.

t/test_repo_consistent.sh
andrii-suse: `excuse` was intentional, meaning that it is the only condition where we shouldn't show failure. Exclude may work here as well though, so you decide

jlausuch (approved):

PR #110, opened 2020-11-27
Disable Factory news on ppc

andrii-suse:
The actual command never succeeded because of folder permission.
It looks the news are not needed to anyone, so just disable it


okurz:
I think the news would be appreciated for all active openSUSE products, i.e. x86_64, aarch64, ppc64le and nowadays even s390x :)

PR #111, opened 2020-11-30
Add function to allow modify the rsync destination

jlausuch:


jlausuch:
In favor of https://github.com/os-autoinst/openqa-trigger-from-obs/pull/112

PR #118, opened 2021-02-02
Use "other" directory as default for assets and map only iso and hdd

jlausuch:
This way, if we want to sync some other file extensions (e.g. txt)
it will end up by default in other. For hdd we can use the same map
we have in create_asset function

### DIFF
--- a/script/cfg.py
+++ b/script/cfg.py
@@ -64,9 +64,9 @@ for flavor in {FLAVORLIST,}; do
         [ ! -z "$src" ] || continue
         dest=$src
         ''' + rsync_fix_dest(distri, version, staging) + '''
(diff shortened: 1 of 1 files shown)


script/cfg.py
```diff
@@ -164,7 +165,7 @@ def rsync_repodir2():
 def openqa_call_fix_destiso(distri, version, staging):
     if not staging:
         return ''
-    if (version == 'Factory' or len(staging)>1):
```
jlausuch, line 167: that was caught automatically by my editor. I can leave it intact if desired.

andrii-suse (approved):

b10n1k (approved):
looks good

PR #121, opened 2021-02-22
ci: Don't install git

andrii-suse:
Without it openssh-server was failing to install:

Error: Subprocess failed. Error: RPM failed: find: '/etc/init.d/rc[35].d': No such file or directory
ERROR: neither useradd nor busybox found!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,12 +3,12 @@ version: 2.1
 jobs:
   unit:
     docker:
-      - image: registry.opensuse.org/opensuse/tumbleweed
+      - image: registry.opensuse.org/opensuse/leap
     steps:
       - checkout
       - run:
           command: |
-            zypper -n in --no-recommends diffutils make python3 tar git-core || :
+            zypper -n in --no-recommends diffutils make python3 tar git-core
       - checkout
       - run: git clean -df
       - run: make test_regen_all
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,14 +3,13 @@ version: 2.1
 jobs:
   unit:
     docker:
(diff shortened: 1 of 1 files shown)


andrii-suse:
Something is wrong in docker inside CircleCI - I cannot reproduce it locally. Let's wait new update and try again:
```
-bash-5.1# ls -la /usr/sbin/useradd
-rwxr-xr-x 1 root root 122800 Feb 11 12:21 /usr/sbin/useradd
-bash-5.1# if [ -f /usr/sbin/useradd ]; then echo 1111; else echo 2222; fi
1111
-bash-5.1# if [ -x /usr/sbin/useradd ]; then echo 1111; else echo 2222; fi
2222
-bash-5.1# test -x /usr/sbin/useradd || echo no
no
-bash-5.1# test -f /usr/sbin/useradd || echo no
-bash-5.1# whoami
root
```

okurz:
I would suggest to stay on Tumbleweed unless it's a temporary workaround for a known, reported issue

andrii-suse:
The way I see the problem: git-core will not install, because its dependency openssh-server fails to install on old kernels (with globc 2.33) when bash 5.1 is in use.
I assume CircleCI must address the problem, because it is widespread, then we will switch back to TW.
We can switch executor type from docker to VM, but I don't think that it worth any effort, because the unit tests have very basic dependencies (core python3, m4, bash), which should behave similarly in every environment.

andrii-suse:
On second thought, we can try skip installing git at all. I clearly remember it didn't work in some cases, but now it just prints `Falling back to CircleCI's native git client but the behavior may be different from official git.` . Which shouldn't be a problem here.

okurz:
cool :)

PR #124, opened 2021-03-03
Make s390x repo name shorter for Leap

SergioAtSUSE:
z/VM parmfile has a length line limit of 72 characters.

- Remove "openSUSE-" prefix to make the name 10 characters shorter
- Dedicate directory to get rid of aarc64-ppc64le-x86_64 in the
directory name


SergioAtSUSE:
https://github.com/os-autoinst/openqa-trigger-from-obs/pull/126

PR #130, opened 2021-04-26
Sync WSL based on leap15.3

mloviska:


mloviska:
@jlausuch

andrii-suse:
The path in dist_path doesn't look correct at the moment - I see 15.3 binaries are at path `openSUSE:Leap:15.3:WSL/standard/x86_64/wsl-appx/`, similar to what defined in `openSUSE:Leap:15.2:WSL.xml`
In case if layout in 15.2 and 15.3 is expected to be the same - we can use single xml to cover both projects. For this we can change `openSUSE:Leap:15.2:WSL.xml` like below:
```
<openQA project_pattern="openSUSE:Leap:(?P&lt;version&gt;15.[2-3]):WSL"
    dist_path="obspublish::openqa/openSUSE:Leap:${version}:WSL/standard/x86_64/wsl-appx">
```
(and maybe rename openSUSE:Leap:15.2:WSL.xml to something like `openSUSE:Leap:15:WSL.xml` . Name of xml doesn't matter - only folder must match name of OBS project to enable automatic sync start)

mloviska:
> The path in dist_path doesn't look correct at the moment - I see 15.3 binaries are at path `openSUSE:Leap:15.3:WSL/standard/x86_64/wsl-appx/`, similar to what defined in `openSUSE:Leap:15.2:WSL.xml`
> In case if layout in 15.2 and 15.3 is expected to be the same - we can use single xml to cover both projects. For this we can change `openSUSE:Leap:15.2:WSL.xml` like below:
> 
> ```
> <openQA project_pattern="openSUSE:Leap:(?P&lt;version&gt;15.[2-3]):WSL"
>     dist_path="obspublish::openqa/openSUSE:Leap:${version}:WSL/standard/x86_64/wsl-appx">
> ```
> 
> (and maybe rename openSUSE:Leap:15.2:WSL.xml to something like `openSUSE:Leap:15:WSL.xml` . Name of xml doesn't matter - only folder must match name of OBS project to enable automatic sync start)

Nice catch Andrii, I have played with the paths and forgot to revert my changes. Thanks!

andrii-suse (approved):

PR #137, opened 2021-07-15
Add support for ISO and HDD in the same flavor

andrii-suse:


script/cfg.py
```diff
@@ -91,8 +91,11 @@ def rsync_commands(checksum):
                     break
                 fi
             done
-            [ -n "$folder" ] || continue
-            echo "rsync --timeout=3600 -tlp4 --specials PRODUCTISOPATH/$folder/$src /var/lib/openqa/factory/hdd/"
+            if [[ $src =~ .iso$ ]]; then
+                echo "rsync --timeout=3600 -tlp4 --specials PRODUCTISOPATH/$folder/$src /var/lib/openqa/factory/iso/"
+            else
+                echo "rsync --timeout=3600 -tlp4 --specials PRODUCTISOPATH/$folder/$src /var/lib/openqa/factory/hdd/"
+            fi
```
okurz, line 98: how about

```shell
[[ $src =~ .iso$ ]] && asset="iso" || asset="hdd"
echo "rsync --timeout=3600 -tlp4 --specials PRODUCTISOPATH/$folder/$src /var/lib/openqa/factory/$asset/"
```

andrii-suse: There are many places like this and I don't think that the difference worth the effort of another CI round. Feel free to tweak such places and submit in another PR, so I will be glad to accept it.
On a bigger scale - I see the whole python code as something which needs to be rewritten in a way that it produces the same .after files from given .xml for each project. I have some ideas how to make it more friendly, but nothing particular yet.

okurz: https://github.com/os-autoinst/openqa-trigger-from-obs/pull/138

PR #139, opened 2021-08-19
Add vagrant_boxes to TW

andrii-suse:


andrii-suse:
It looks this is not needed at the moment

PR #142, opened 2021-10-13
Use OBS built container as CI base to provide all dependencies

okurz:


okurz:
> The container has the same problem as the manual `zypper in …` before. Maybe `devel:openQA:Leap:15.3` can be added to the repository paths for building the container.

well, as an alternative we could try that. But I think a separate subrepo with just the minimum required would be cleaner

okurz:
Updated to use Tumbleweed as base so that we don't need to care to bump the OS version number. Adjusting the OBS service project accordingly

Martchus (dismissed):

okurz:
What's "invalid reference format" https://app.circleci.com/pipelines/github/os-autoinst/openqa-trigger-from-obs/1044/workflows/e0beeda9-7a0e-48a1-ad0b-42d8519d8b36/jobs/2351 ?

okurz:
All done. After merge I would setup the github-OBS webhook connection to automatically trigger the dependency container build on every change in this repo

container/Dockerfile
```diff
@@ -0,0 +1,4 @@
+#!BuildTag: openqa-trigger-from-obs-dev
+#!UseOBSRepositories
+FROM opensuse/tumbleweed
+RUN zypper -n in --no-recommends diffutils tar make git-core python3-black python3-importlib-metadata
```
b10n1k, line 4: are you sure you need `python3-importlib-metadata` because I didnt find any reference in https://github.com/os-autoinst/openqa-trigger-from-obs/blob/master/script/scriptgen.py? 
Then I tried local pulling the imageand run the test

```c314f5120a73:/openqa-trigger-from-obs # zypper -n in --no-recommends make python3-black
Loading repository data...
Reading installed packages...
'make' is already installed.
No update candidate for 'make-4.4.1-3.3.x86_64'. The highest available version is already installed.
'python3-black' not found in package names. Trying capabilities.
Resolving package dependencies...

The following NEW package is going to be installed:
  python313-black

1 new package to install.

Package download size:   446.3 KiB

Package install size change:
              |       2.0 MiB  required by packages that will be installed
     2.0 MiB  |  -      0 B    released by packages that will be removed

Backend:  classic_rpmtrans
Continue? [y/n/v/...? shows all options] (y): y
Preloading: python313-black-25.1.0-4.1.noarch.rpm [done]
Preload finished. [success (13.7 KiB/s) ] ..............................................................................................................................................................................................................................[done]
Retrieving: python313-black-25.1.0-4.1.noarch (openSUSE-Tumbleweed-Oss)                                                                                                                                                                                   (1/1), 446.3 KiB    

Checking for file conflicts: ...........................................................................................................................................................................................................................................[done]
(1/1) Installing: python313-black-25.1.0-4.1.noarch ....................................................................................................................................................................................................................[done]
Running post-transaction scripts .......................................................................................................................................................................................................................................[done]
c314f5120a73:/openqa-trigger-from-obs # make test_python_style
black --check --fast --diff script/scriptgen.py 
All done! ✨ 🍰 ✨
1 file would be left unchanged.
c314f5120a73:/openqa-trigger-from-obs # zypper se -i python3-importlib-metadata
Loading repository data...
Reading installed packages...
No matching items found.
```

d3flex (changes requested):
Maybe I am wrong and maybe there is another reasoning for python3-importlib-metadata. Would you mind to double check?

d3flex (approved):

PR #143, opened 2021-11-02
Add openSUSE:Leap:15.4:Images, enable VBox image for 15.3

Vogtinator:
See commits.

The version of JeOS in 15.4 hasn't been updated yet, so the filenames still say "15.3", but syncing should still work once that's fixed in the package.


andrii-suse:
the integration test fails because of apparmor : `/usr/bin/perl: error while loading shared libraries: libm.so.6: cannot stat shared object: Permission denied` . Maybe you know an easy fix for it?

andrii-suse:
I mean the CI shouldn't block this PR, so I am good to merge it if it is urgent

Vogtinator:
> the integration test fails because of apparmor : `/usr/bin/perl: error while loading shared libraries: libm.so.6: cannot stat shared object: Permission denied` . Maybe you know an easy fix for it?

Where do you see that message?

This sounds like either usrmerge gone wrong, glibc incompatibility or somehow a partial upgrade.

Martchus:
Possibly just https://lists.opensuse.org/archives/list/factory@lists.opensuse.org/thread/P5ATW7DDZ56ZEY6KNT6CLLQIWNNGZYTC/ ?

At least

```
terminate called after throwing an instance of 'std::system_error'
  what():  Operation not permitted
```

looks much like it.

andrii-suse:
> Where do you see that message?

I've added code that cats the logs in https://github.com/os-autoinst/openqa-trigger-from-obs/pull/144

https://app.circleci.com/pipelines/github/os-autoinst/openqa-trigger-from-obs/450/workflows/97aaae98-e2e4-4084-9281-af9ecad76b10/jobs/775?invite=true#step-102-168

Maybe it is some incompatibility between host (ubuntu) and leap docker. Probably we have to disable integration tests which use systemctl, shame because they were good to test approfile until recent changes

andrii-suse:
I've merged #144, so if you rebase - all should be green

Vogtinator:
Done

PR #145, opened 2021-11-12
Add Containers flavor for Leap images

jlausuch:
This allows creating an additional isos post with specific flavor
to validate container images without reusing JeOS or any other.

### DIFF
--- a/script/cfg.py
+++ b/script/cfg.py
@@ -213,6 +213,8 @@ for flavor in {FLAVORALIASLIST,}; do
         ''' + openqa_call_start_distri(flavor_distri) + '''
         [[ ! -v flavor_filter[@] ]] || [ -z "${flavor_filter[$flavor]}" ] || filter=${flavor_filter[$flavor]}
         [ $filter != Appliance ] || filter="qcow2"
+        version=VERSIONVALUE
+        if [ -z "${norsync_filter[$flavor]}" ] || [ -z $build1 ]; then {
         iso=$(grep "$filter" __envsub/files_iso.lst 2>/dev/null | grep $arch | head -n 1)
         ''' + openqa_call_start_fix_iso(archs) + '''
         build=$(echo $iso | grep -o -E '(Build|Snapshot)[^-]*' | grep -o -E '[0-9]\.?[0-9]+(\.[0-9]+)*' | tail -n 1) || :
@@ -226,13 +228,14 @@ for flavor in {FLAVORALIASLIST,}; do
         buildex=${buildex/.qcow2/}
         build1=$build
         destiso=$iso
-        version=VERSIONVALUE
         [ -z "__STAGING" ] || build1=__STAGING.$build
         ''' + openqa_call_fix_destiso(distri, version, staging) + '''
         repo0folder=${destiso%.iso}
         ''' + (repo0folder if repo0folder else "") + '''
         [ "$arch" != . ] || arch=x86_64
         ''' + openqa_call_news(news, news_archs) + '''
+        }
+        fi
         echo "/usr/bin/openqa-cli api -X post isos \\\\\"
 (
  echo \" DISTRI=$distri \\\\
--- a/script/scriptgen.py
+++ b/script/scriptgen.py
@@ -154,6 +154,7 @@ class ActionBatch:
         self.asset_folders = {}
         self.asset_tags = {}
         self.asset_rsync = {}
+        self.norsync = {}
         self.isos = []
         self.iso_folder = {}
         self.iso_5 = ""
@@ -319,6 +320,8 @@ class ActionBatch:
                     self.iso_5 = node.attrib.get("iso_5")
                 if node.attrib.get("fixed_iso", ""):
                     self.fixed_iso = node.attrib["fixed_iso"]
+                if node.attrib.get("rsync", "1") == "0":
+                    self.norsync[f] = 1
 
         if node.attrib.get("iso", "") and node.attrib.get("name", ""):
             for iso in node.attrib["name"].split("|"):
@@ -669,6 +672,8 @@ class ActionBatch:
 
     def gen_print_array_no_rsync(self, f):
         self.p("declare -A norsync_filter", f)
+        for fl in self.norsync:
+            self.p("norsync_filter[{}]='{}'".format(fl, 1), f)
         for fl, h in zip(self.flavors, self.assets):
             if self.asset_rsync.get(h, 1) == "0":
                 self.p("norsync_filter[{}]='{}'".format(h, 1), f)
@@ -972,7 +977,7 @@ done < <(sort __envsub/files_asset.lst)""",
                 + self.assets_flavor
                 + """ ]] """
                 + arch_expression
-                + """ || while read src; do
+                + """ || [[ ${norsync_filter[$filter]} == 1 ]] || while read src; do
      echo " ASSET_$i=$src \\\\"
      asset_tag=""
      for mask in "${!asset_tags[@]}"; do
--- a/t/obs/openSUSE:Leap:15.2:Images:ToTest/print_openqa.before
+++ b/t/obs/openSUSE:Leap:15.2:Images:ToTest/print_openqa.before
@@ -54,3 +54,11 @@
  VERSION=15.2 \
  _OBSOLETE=1
 
+/usr/bin/openqa-cli api -X post isos \
+ ARCH=x86_64 \
+ BUILD=20.3 \
+ DISTRI=opensuse \
+ FLAVOR=Container-Image \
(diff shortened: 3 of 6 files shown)


jlausuch:
@Vogtinator @andrii-suse Please have a look and let me know if you agree.
My goal is to have something similar to this:

Screenshot 2021-11-12 at 18 49 08

andrii-suse:
Since no additional content is generated in t/ folder - no change is expected.
Normally `isos post` command needs an image, which will be loaded (e.g. iso or hdd). And each flavor in the xml refers at least one of such image. E.g. flavor `Vagrant` will test `libvirt.box` and `virtualbox.box`.
Which images will be tested in the Container-Image flavor? I think it will be easier to define which new `isos post` (aka `api -X post isos`) is required for the new flavor, then think how to define XML so new eventual artifacts are included into rsync and the new `isos post` is generated properly.

jlausuch:
> Since no additional content is generated in t/ folder - no change is expected.
Normally isos post command needs an image, which will be loaded (e.g. iso or hdd).

Yeah, I know, but wouldn't this generate another isos post with no `HDD_1` or `ISO` and that flavor?

> Which images will be tested in the Container-Image flavor?

Some 3rd party images defined in the jobs (CentOS and Ubuntu HDD_1) to test container images on top, so that's why I need to define this flavor for those special images. 
 
> I think it will be easier to define which new `isos post` (aka `api -X post isos`) is required for the new flavor, then think how to define XML so new eventual artifacts are included into rsync and the new `isos post` is generated properly.

The post would be something simple, without any HDD or ISO defined:

```
/usr/share/openqa/script/client isos post --host localhost \
 ARCH=$arch \
 BUILD=$build \
 DISTRI=opensuse \
 FLAVOR=Container-Image \
 VERSION=15.3 
```

andrii-suse:
> Some 3rd party bootable images defined in the jobs (CentOS and Ubuntu HDD_1) to test openSUSE container images on top, so that's why I need to define this flavor for those special cases.

Just to understand it better :  how `BUILD=$build` from `isos post` command will influence the outcome? Since testing in openQA happens before `ToTest` project get published - you probably will not be able to access any artifacts from that $build, so it is not clear how the job outcome will relate to the snapshot referenced by $build.

jlausuch:
> > Some 3rd party bootable images defined in the jobs (CentOS and Ubuntu HDD_1) to test openSUSE container images on top, so that's why I need to define this flavor for those special cases.
> 
> Just to understand it better : how `BUILD=$build` from `isos post` command will influence the outcome? Since testing in openQA happens before `ToTest` project get published - the job probably will not be able to access any artifacts from that $build, so it is not clear how the job outcome will relate to the snapshot referenced by $build.

We pull `registry.opensuse.org/opensuse/leap/15.3/images/totest/containers/opensuse/leap:15.3` (latest) container image, if tests are ok, we publish it to `registry.opensuse.org/opensuse/leap:15.3`. This is already happening, so I'm not changing this logic. I need the same build as the other images (JeOS, Vagrant, etc) because we use the same job/build to validate all the images at the same time (including the container image). Check [this link for Leap 15.3 images in O3](https://openqa.opensuse.org/tests/overview?distri=opensuse&version=15.3&build=9.279&groupid=77&flavor=JeOS-for-kvm-and-xen&test=containers_docker_on_15.2_gm&&test=containers_podman_on_15.2_gm&&test=containers_on_centos_host_x86_64&
&test=containers_on_ubuntu_host_x86_64)
Currently, these jobs are under `JeOS-for-kvm-and-xen` flavor, which is wrong, because they are not using any JeOS image, they boot images that are not part of that build/version (centos, ubuntu, leap 15.2).

Vogtinator:
Currently this is a noop, because the flavor doesn't include any files to sync.

So far this wasn't implemented because nothing needs to be synced for container testing. It might work to use the `.containerinfo` file as a flag though, but I never tried that. It might need some changes to the various scripts involved.

One advantage of that approach is that the build number of the container image is then available to the openQA test, so it could use that specific revision. In practice this only causes restarts of older runs to fail though, because there's always just a single container image build in `:ToTest`.

> > Some 3rd party bootable images defined in the jobs (CentOS and Ubuntu HDD_1) to test openSUSE container images on top, so that's why I need to define this flavor for those special cases.
> 
> Just to understand it better : how `BUILD=$build` from `isos post` command will influence the outcome? Since testing in openQA happens before `ToTest` project get published - the job probably will not be able to access any artifacts from that $build, so it is not clear how the job outcome will relate to the snapshot referenced by $build.

The containers repo of `:ToTest` is special - it's publish enabled to allow downloads from the registry. For publishing, the container binaries are released into `openSUSE:Containers:...`.

jlausuch:
Good point. I can get `opensuse-leap-image.x86_64-15.2.0-docker-Build8.284.docker.tar.registry.txt` as asset, I've done that before for SLE containers. However, I would like to keep the same build number as for the other images, otherwise the tests won't be in the same build group. 
https://download.opensuse.org/repositories/openSUSE:/Leap:/15.2:/Images:/ToTest/containers/

jlausuch:
Looks good! Thanks.

I'm sorry, I thought the feature `rsync='0'` was already implemented. It has been used in the past for OSD:
https://gitlab.suse.de/openqa/openqa-trigger-from-ibs-plugin/-/blob/658015525bdc093ddbb86c29d254a42c57ec419d/xml/SUSE:SLE-15:Update:CR:Images.xml#L6

andrii-suse:
It was implemented only for asset tag. But I agree that this generation of bash which prints bash is way too complex to follow. I wish somebody had an idea how to easier achieve the same testability and ability to preview all the actions for all projects for every change in definition.

andrii-suse (approved):

jlausuch:
@Vogtinator do we want to sync the asset `opensuse-leap-image.x86_64-15.2.0-docker-Build8.284.docker.tar.registry.txt` to check which container image build we are testing or we can live without it?

```
    <flavor name="Container-Image">
        <asset filemask='opensuse-leap-image.*.txt' rsync='0'/>
    </flavor>
```

jlausuch:
Shall we merge this then?

ggardet:
Please take care of aarch64 as well.

jlausuch:
> Please take care of aarch64 as well.

that has not changed. For Leap 15.2 there are no aarch64 jobs currently, e.g. https://openqa.opensuse.org/tests/overview?distri=opensuse&version=15.2&build=31.596&groupid=65
for 15.3 there are. https://openqa.opensuse.org/tests/overview?distri=opensuse&version=15.3&build=9.281&groupid=77

Vogtinator:
> > Please take care of aarch64 as well.
> 
> that has not changed. For Leap 15.2 there are no aarch64 jobs currently, e.g. https://openqa.opensuse.org/tests/overview?distri=opensuse&version=15.2&build=31.596&groupid=65 for 15.3 there are. https://openqa.opensuse.org/tests/overview?distri=opensuse&version=15.3&build=9.281&groupid=77

There is https://openqa.opensuse.org/tests/overview?distri=opensuse&version=15.2&build=1.288&groupid=66, which also takes care of container images. There's no `jeos-containers` test inside, which could be fixed, but might not be worth it considering the close EOL.

jlausuch:
> > > Please take care of aarch64 as well.
> > 
> > 
> > that has not changed. For Leap 15.2 there are no aarch64 jobs currently, e.g. https://openqa.opensuse.org/tests/overview?distri=opensuse&version=15.2&build=31.596&groupid=65 for 15.3 there are. https://openqa.opensuse.org/tests/overview?distri=opensuse&version=15.3&build=9.281&groupid=77
> 
> There is https://openqa.opensuse.org/tests/overview?distri=opensuse&version=15.2&build=1.288&groupid=66, which also takes care of container images. There's no `jeos-containers` test inside, which could be fixed, but might not be worth it considering the close EOL.

I would proceed with the current state which fixes the urgent failures, and proceed with adding more coverlage later on if needed. 15.4 will also need coverage in the near future.

ggardet:
> > > Please take care of aarch64 as well.
> > 
> > 
> > that has not changed. For Leap 15.2 there are no aarch64 jobs currently, e.g. https://openqa.opensuse.org/tests/overview?distri=opensuse&version=15.2&build=31.596&groupid=65 for 15.3 there are. https://openqa.opensuse.org/tests/overview?distri=opensuse&version=15.3&build=9.281&groupid=77
> 
> There is https://openqa.opensuse.org/tests/overview?distri=opensuse&version=15.2&build=1.288&groupid=66, which also takes care of container images. There's no `jeos-containers` test inside, which could be fixed, but might not be worth it considering the close EOL.

Indeed, 15.2 is EOL in few weeks (days?), so let 15.2 aside.

PR #148, opened 2021-12-08
Add Container-Image flavor to Leap 15.4

jlausuch:
Same as for Leap 15.2 and 15.3


jlausuch:
@Vogtinator @andrii-suse

jlausuch:
I'm trying to add the tests, but I get
```
rsync -4 --list-only 'obspublish::openqa/openSUSE:Leap:15.4:Images:ToTest/images/*/kiwi-templates-JeOS:kvm-and-xen/'
rsync: getaddrinfo: obspublish 873: Name or service not known
rsync error: error in socket IO (code 10) at clientserver.c(127) [Receiver=3.1.3]
```

andrii-suse:
So the tests do generate `rsync` and `isos post` commands based on xml to make sure you acknowledge all the changes and side effects on all projects. In this particular case the CI complains that the new isos post commands were not added to print_openqa.before - can you see the error in CI ?
```
F-obs/openSUSE:Leap:15.4:Images:ToTest test_before_after_diff.sh:FAIL obs/openSUSE:Leap:15.4:Images:ToTest print_openqa : --- obs/openSUSE:Leap:15.4:Images:ToTest/print_openqa.before	2021-12-08 10:40:38.870208543 +0000
+++ obs/openSUSE:Leap:15.4:Images:ToTest/print_openqa.after	2021-12-08 10:41:24.234904520 +0000
@@ -141,3 +141,19 @@
  VERSION=15.4 \
  _OBSOLETE=1
 
+/usr/bin/openqa-cli api -X post isos \
+ ARCH=aarch64 \
+ BUILD=4.9 \
+ DISTRI=opensuse \
+ FLAVOR=Container-Image \
+ VERSION=15.4 \
+ _OBSOLETE=1
```

You don't need to run rsync commands to add those - you just needed to regenerate .before files, e.g. `Update .before scripts in test folders with make test_update_before_files`  ( https://github.com/os-autoinst/openqa-trigger-from-obs#changes-needed-for-existing-projects ). Or maybe just add the new `isos post` commands from the CI output to `obs/openSUSE:Leap:15.4:Images:ToTest/print_openqa..before`

With such approach (in general case) :  the pull request will show all changes that will be put into production and reviewers will make sure that nothing else breaks. So the CI is this strict to ensure that no unwanted surprises happens after deployment

jlausuch:
> So the tests do generate `rsync` and `isos post` commands based on xml to make sure you acknowledge all the changes and side effects on all projects. In this particular case the CI complains that the new isos post commands were not added to print_openqa.before - can you see the error in CI ?
> 
> ```
> F-obs/openSUSE:Leap:15.4:Images:ToTest test_before_after_diff.sh:FAIL obs/openSUSE:Leap:15.4:Images:ToTest print_openqa : --- obs/openSUSE:Leap:15.4:Images:ToTest/print_openqa.before	2021-12-08 10:40:38.870208543 +0000
> +++ obs/openSUSE:Leap:15.4:Images:ToTest/print_openqa.after	2021-12-08 10:41:24.234904520 +0000
> @@ -141,3 +141,19 @@
>   VERSION=15.4 \
>   _OBSOLETE=1
>  
> +/usr/bin/openqa-cli api -X post isos \
> + ARCH=aarch64 \
> + BUILD=4.9 \
> + DISTRI=opensuse \
> + FLAVOR=Container-Image \
> + VERSION=15.4 \
> + _OBSOLETE=1
> ```
> 
> You don't need to run rsync commands to add those - you just needed to regenerate .before files, e.g. `Update .before scripts in test folders with make test_update_before_files` ( https://github.com/os-autoinst/openqa-trigger-from-obs#changes-needed-for-existing-projects ). Or maybe just add the new `isos post` commands from the CI output to `obs/openSUSE:Leap:15.4:Images:ToTest/print_openqa.before`
> 
> With such approach (in general case) : the pull request will show all changes that will be put into production and reviewers will make sure that nothing else breaks. So the CI is this strict to ensure that no unwanted surprises happens after deployment

Yeah, I know about all this, I've done it dozens of times for IBS plugin :)
Actually, I run before `make test_regen_all`, and then `bash -x read_files.sh` and that gives me the output that of my previous comment.

t/obs/openSUSE:Leap:15.4:Images:ToTest/print_openqa.before
```diff
@@ -1,4 +1,4 @@
-/usr/bin/openqa-cli api -X post isos \
+/usr/share/openqa/script/client isos post --host localhost \
  ARCH=x86_64 \
  ASSET_256=openSUSE-Leap-15.3-JeOS.x86_64-15.3-kvm-and-xen-Build4.9.qcow2.sha256 \
```
jlausuch, line 3: @Vogtinator btw, this is suspicious... have we properly updated the version in JeOS images for 15.4?

andrii-suse:
> Yeah, I know about all this, I've done it dozens of times for IBS plugin :)

Ah good then, ibs has simpler networking, so it is easier to reach IBS backend. Unfortunately it is much harder to reach the list of files in OBS, so it may work only if you have obspublush mudule for rsync configured - I usually run it on the o3 host itself as geekotest user and the password exported with RSYNC_PASSWORD global variable. Fabian had a nice idea how to use OBS API to do the same, which probably worth to do, but it complicates thing in different dimension (like osc dependency and using different tools for collecting list of files and syncing).

t/obs/openSUSE:Leap:15.4:Images:ToTest/print_openqa.before
andrii-suse: git sometimes behaves weird - not sure how the old command was used here. 
If I search in github - I cannot find any reference to `/usr/share/openqa/script/client` 
I'd just get new `git clone` of openqa-trigger-from-obs and retry this PR with it. Otherwise I can do the same and see if it gets correct this time.

Vogtinator: The JeOS images have not been updated yet - waiting for the Kiwi update to land in 15.4.

andrii-suse (approved):
In general I would keep old the .lst files, or maybe updated them in separate SR, (because it is easier to review what actually changed in generation from .lst to .before), but LGTM

Vogtinator:
> Fabian had a nice idea how to use OBS API to do the same, which probably worth to do, but it complicates thing in different dimension (like osc dependency and using different tools for collecting list of files and syncing).

osc isn't really necessary, it's just plain HTTP GET with some trivial XML output. rsync would still be required though, so some duplication (and possible inconsistency) is possible.

PR #151, opened 2022-01-19
Add KDE:Medias for syncing Argon and Krypton

Vogtinator:
Currently those are handled by jenkins.qa.suse.de.

One downside of adding them here is that the `VERSION` value is no longer derived from the ISO name after syncing, but for Krypton that is currently overridden by the medium type anyway.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,7 +20,7 @@ jobs:
       - checkout
       - run:
           command: |
-            zypper -n ar https://download.opensuse.org/repositories/devel:/languages:/python:/backports/openSUSE_Leap_15.3 py
+            zypper -n ar https://download.opensuse.org/repositories/devel:/languages:/python:/backports/15.3 py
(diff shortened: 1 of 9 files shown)


andrii-suse:
The only problem here is that botmaster will not trigger syncs automatically - iirc currently it does only for prefixes defined at https://github.com/openSUSE/openSUSE-release-tools/blob/3990fecd037e4a08735b36a0d1cacf3cc5b7ac04/gocd/monitors.gocd.yaml#L124

Vogtinator:
> The only problem here is that botmaster will not trigger syncs automatically - iirc currently it does only for prefixes defined at https://github.com/openSUSE/openSUSE-release-tools/blob/3990fecd037e4a08735b36a0d1cacf3cc5b7ac04/gocd/monitors.gocd.yaml#L124

Thanks for the hint, I'll open a PR

andrii-suse (approved):

PR #152, opened 2022-02-14
Add retry to openqa-cli

andrii-suse:


andrii-suse:
Actually retrying is even worse, because the jobs are actually started despite the proxy error and retry may still fail because of Apache timeout. Should better use async requests, it looks openQA should allow it

PR #154, opened 2022-02-17
Add wicked-debug* packages

cfconrad:
Because of https://progress.opensuse.org/issues/106963 I would like to
sync the `wicked-debug*` packages to o3.


cfconrad:
This is what I have done, I do *not* have deep knowledge of this...

```bash
grep -rl mraa-debug\* | xargs -IXXX sed -i 's/mraa-debug\*/mraa-debug*,wicked-debug*/' XXX
grep -rl  'include=mraa-debug\*,wicked-debug\*' | xargs -IXXX sed -i 's/include=mraa-debug\*,wicked-debug\*/include=mraa-debug* --include=wicked-debug*/' XXX
```

andrii-suse:
> This is what I have done, I do _not_ have deep knowledge of this...

If the tests are green - then all should be good. For next time: `make test_update_before_files` should do the same.

PR #155, opened 2022-02-20
t: Enable rest integration tests

andrii-suse:


t/docker/lib/common.sh
```diff
@@ -23,14 +23,17 @@ touch /var/lib/openqa/osc-plugin-factory/factory-package-news/factory-package-ne
...
 systemctl restart openqa-webui.service
 systemctl restart openqa-websockets.service
+systemctl restart rsyncd
 
 [ "$METHOD" != rest ] || systemctl restart openqa-gru.service
 
 # wait for webui to become available
 sleep 2
-attempts_left=10
+attempts_left=25 # sometimes it takes 1 min for starting service (when assests download is slow?)
```
okurz, line 36: s/assests/assets/

t/docker/script/openSUSE:Leap:15.2:Staging:A.sh
```diff
@@ -15,9 +15,9 @@ set -x
 # make sure run did happen
 test -f /var/lib/openqa/factory/iso/openSUSE-Leap-15.2-Staging:A-Staging-DVD-x86_64-Build248.1-Media.iso
 test -f /opt/openqa-trigger-from-obs/openSUSE:Leap:15.2:Staging:A/.run_last/openqa.cmd.log
-grep -q 'scheduled_product_id => 1' /opt/openqa-trigger-from-obs/openSUSE:Leap:15.2:Staging:A/.run_last/openqa.cmd.log || cat /opt/openqa-trigger-from-obs/openSUSE:Leap:15.2:Staging:A/.run_last/*.cmd.log
+grep -q '"scheduled_product_id":1' /opt/openqa-trigger-from-obs/openSUSE:Leap:15.2:Staging:A/.run_last/openqa.cmd.log || cat /opt/openqa-trigger-from-obs/openSUSE:Leap:15.2:Staging:A/.run_last/*.cmd.log
```
okurz, line 18: seems like these tests haven't been called at all after switching to openqa-cli?

andrii-suse: Yeah it looks the script was somehow exiting when function `prepare_project` exits without any error, just silently stop executing commands:
https://app.circleci.com/pipelines/github/os-autoinst/openqa-trigger-from-obs/480/workflows/01e80f46-1b39-440c-87cb-fc7d9c920aa7/jobs/867?invite=true#step-102-235
These are corresponding places in code:
https://github.com/os-autoinst/openqa-trigger-from-obs/blob/master/t/docker/script/openSUSE:Leap:15.2:Staging:A.sh#L12
https://github.com/os-autoinst/openqa-trigger-from-obs/blob/master/t/docker/lib/common.sh#L72
It is another mystery, especially that I could reproduce the same locally. Needs some more investigation.
(I've double checked - it worked properly in this PR though)

PR #156, opened 2022-02-22
Use async flag in isos post

andrii-suse:


okurz (commented):
This just triggers tests but doesn't check results so it can be that we end up with unnoticed missed or incomplete test sets

andrii-suse:
My understanding:

- It is task for other tools to review outcome of testing. 
- Release Managers shouldn't be required to look to sync status in addition.

So, even if the gru failed to start some tests - the sync is complete from ObsRsync point of view and release tools should review that all required tests are started and have passed properly.

okurz:
> My understanding:
> 
>     * It is task for other tools to review outcome of testing.
> 
>     * Release Managers shouldn't be required to look to sync status in addition.
> 
> 
> So, even if the gru failed to start some tests - the sync is complete from ObsRsync point of view and release tools should review that all required tests are started and have passed properly. Showing failure in both places will not help much and will only create unneeded noise. I see it as kind of `single responsibility` here - the sync does only what is defined in xml; other components should keep responsibility for verifying outcome and completeness of tests started.

Sure. I would only be concerned with anything actually falls in the scope of openqa-trigger-from-obs then. For example let's say the API of openQA changes and openqa-trigger-from-obs fails. It would be good if the tooling would know about that.

andrii-suse:
Currently if any command during sync fails - the gru job (`obs_rsync_run`) will be in `failed` state, plus column `Last failure` in the table at `/admin/obs_rsync` references to it and the details should be picked up in the logs. 
So if API changed - I expect `isos post?async=1` will return an error and it will be reflected.
Not sure what kind of additional indication will be helpful.

Martchus:
> So if API changed - I expect isos post?async=1 will return an error and it will be reflected as described above.

I'd expect the same. It was already bad if we'd break backwards compatibility but breaking it in a non-detectable way would be even worse (so I suppose it won't happen).

PR #157, opened 2022-03-07
Add JeOS and MicroOS images to openSUSE:Factory:Staging

Vogtinator:
Unfortunately it appears like adding JeOS now made it ignore `Tumbleweed-DVD`. It's still part of `read_files.sh` as well as `files_iso.lst`, but not in any other files. @andrii-suse any idea? It might work for `openSUSE:Factory:ToTest` because it wraps everything in `<batch/>`.


Vogtinator:
Also it looks like the `FLAVOR` is wrong. That appears to be coming from some hardcoded place...

coolo:
HDD_1=openSUSE-Tumbleweed-JeOS.x86_64-15.1.0-kvm-and-xen-Build30.39.qcow2 won't do - we need some unique component in there (i.e. a rename must happen)

coolo:
And the openQAMapping is `s/openSUSE-/openSUSE-Staging:$LETTER-/` - so it needs to be a very specific renaming

andrii-suse:
> HDD_1=openSUSE-Tumbleweed-JeOS.x86_64-15.1.0-kvm-and-xen-Build30.39.qcow2 won't do - we need some unique component in there (i.e. a rename must happen)

So currently it is searching for `FLAVOR` inside name of iso/qcow2 file and injects `Staging` there. If flavor would be just `kvm-and-xen` , then if will work properly, otherwise I need to think about other ways.

Vogtinator:
We could use a `kiwi_post_run` script for renaming, but doing it like for DVDs is probably better.

> > HDD_1=openSUSE-Tumbleweed-JeOS.x86_64-15.1.0-kvm-and-xen-Build30.39.qcow2 won't do - we need some unique component in there (i.e. a rename must happen)
> 
> So currently it is searching for `FLAVOR` inside name of iso/qcow2 file and injects `Staging` there. If flavor would be just `kvm-and-xen` , then if will work properly, otherwise I need to think about other ways.

Using just `JeOS` as flavor would work too, but for MicroOS there isn't really a good alternative. So we'd need some better way. Maybe just prepend `Staging:B-` to the files?

Vogtinator:
> @Vogtinator it must be some ugly hack I did introduce at some point. I've checked and it works properly if you move the new flavor to be the first in the list (above DVDs). Otherwise I need some time to do some kind of rework there.

That works indeed.

Is there some way to have the TW DVD added as `ISO_1` to the tests triggered by the HDD? That way it can be used as repository.

andrii-suse:
> Using just `JeOS` as flavor would work too, but for MicroOS there isn't really a good alternative. So we'd need some better way. Maybe just prepend `Staging:B-` to the files?

Let's have flavor `JeOS-kvm-and-xen` and use 

> Is there some way to have the TW DVD added as `ISO_1` to the tests triggered by the HDD? That way it can be used as repository.

Not at the moment, there were no such cases yet. But give me some time, I will come with proposal for this and injecting `Staging`.

andrii-suse:
It is already out of control and becomes even worse, but this commit does the job as far as I can see. The only remaining issue is different BUILD in `isos post` commands for DVDs vs qcow2
https://github.com/os-autoinst/openqa-trigger-from-obs/commit/f9f455ceb709  (I couldn't push to your fork)

Vogtinator:
Staging:B at least is ready, the others should soon be too. So next week this could in theory be tried out. There aren't any tests hooked up to the new products yet, that's something for after the syncing works.

andrii-suse (approved):
Looks good, the only problem I see is that `isos post` uses different BUILD variable for each call:  B.3.3 , B.56.8 , B.545.4 
I am not sure how bad it is. Theoretically we can try to concatenate that and use something like `BUILD=B.3.3_56.8_545.4` for each call , or maybe new variable `BUILD_AGG=B.3.3_56.8_545.4`. Not sure if it is a good idea and how hard it will be to achieve and what is the best to proceed here

Vogtinator:
> Looks good, the only problem I see is that `isos post` uses different BUILD variable for each call: B.3.3 , B.56.8 , B.545.4 I am not sure how bad it is. Theoretically we can try to concatenate that and use something like `BUILD=B.3.3_56.8_545.4` for each call , or maybe new variable `BUILD_AGG=B.3.3_56.8_545.4`. Not sure if it is a good idea and how hard it will be to achieve and what is the best to proceed here

I'm not entirely sure either. From an openQA -> OBS tooling perspective it shouldn't matter, but I'm not sure what happens if just one of those products gets a new build. Will it get rescheduled on its own?

andrii-suse:
> I'm not entirely sure either. From an openQA -> OBS tooling perspective it shouldn't matter, but I'm not sure what happens if just one of those products gets a new build. Will it get rescheduled on its own?

The automatic sync happens when rabbit sends notification for OBS project that matches the folder in openqa-trigger-from-obs , i.e. `openSUSE:Factory:Rings:1-MinimalX` and e.g. `openSUSE:Factory:Staging:B` . Then:
-  `read_files.sh` is called, which generates files_iso.lst 
- newly generated files_iso.lst compares against the same file in .run_last/files_iso.lst - and the sync starts if there are any differences

I guess it is proper behavior as long as nobody has better understanding how the workflow should happen.

DimStar77:
This seems to be the PR that blocks Stagings.

currently I have Staging:C and :I 'ready to sync to QA' - but neither ends up there.

Staging:I has in the rsync logs:

```
+ rsync --timeout=3600 -tlp4 --specials 'obspublish-stage::openqa/openSUSE:Factory:Staging:I/images/x86_64/*product*/kiwi-templates-Minimal:kvm-and-xen/*openSUSE-Tumbleweed-Minimal-VM.x86_64-1.0.0-kvm-and-xen-Build3.1.qcow2' /var/lib/openqa/factory/hdd/openSUSE-Staging:I-Tumbleweed-Minimal-VM.x86_64-1.0.0-kvm-and-xen-Build3.1.qcow2
rsync: [sender] change_dir "/openSUSE:Factory:Staging:I/images/x86_64/*product*/kiwi-templates-Minimal:kvm-and-xen" (in openqa) failed: No such file or directory (2)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1674) [Receiver=3.1.3]
```

DimStar77:
> This seems to be the PR that blocks Stagings.
> 
> currently I have Staging:C and :I 'ready to sync to QA' - but neither ends up there.
> 
> Staging:I has in the rsync logs:
> 
> ```
> + rsync --timeout=3600 -tlp4 --specials 'obspublish-stage::openqa/openSUSE:Factory:Staging:I/images/x86_64/*product*/kiwi-templates-Minimal:kvm-and-xen/*openSUSE-Tumbleweed-Minimal-VM.x86_64-1.0.0-kvm-and-xen-Build3.1.qcow2' /var/lib/openqa/factory/hdd/openSUSE-Staging:I-Tumbleweed-Minimal-VM.x86_64-1.0.0-kvm-and-xen-Build3.1.qcow2
> rsync: [sender] change_dir "/openSUSE:Factory:Staging:I/images/x86_64/*product*/kiwi-templates-Minimal:kvm-and-xen" (in openqa) failed: No such file or directory (2)
> rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1674) [Receiver=3.1.3]
> ```

The problem seems to be that the kiwi-template-* is not built as part of 000product as a sub flavor (000product:flavor matching `*product` ) this then runs into this sync error and nothing happens

PR #159, opened 2022-03-22
Leap Micro 5.2: use correct target for images and ISOs

jlausuch:


jlausuch:
@andrii-suse please have a look, since we have no ISOs in one path and Images in another path, I didn't know how to merge all together, so I created 2 different projects.

andrii-suse:
The main question is when all isos / appliances must be synced and the tests triggered. Current model is that name of folder matches OBS project and when that project triggers corresponding rabbitmq event (and the api shows `(un)published` without `dirty` = the sync starts ).
If  one of the projects can be used for such checks - it should be possible to merge the things into single xml . Just let me know. Otherwise I've approved the PR.

andrii-suse (approved):

jlausuch:
Answer from Max:
```
we should treat them separately for that 2 projects, we don't need to trigger them at the same time. We do have the similar setup for openSUSE:Leap:15.4/openSUSE:Leap:15.4:Images, they're belongs to 2 JobGroup on openQA for isos and images.
```

PR #160, opened 2022-04-04
Add sync for D-Installer

jknphy:
Additional PR for monitoring: https://github.com/openSUSE/openSUSE-release-tools/pull/2792


andrii-suse:
Having folder in `t` will make sure that nothing will break after any refactoring. But if you don't want such security, then the .lst file is not needed as well (it used is only in tests)

andrii-suse (approved):

jknphy:
I worked fine: https://openqa.opensuse.org/tests/overview?distri=d-installer&version=&build=3.2&groupid=96

Martchus (approved):

PR #164, opened 2022-05-05
Download selfinstall iso

mloviska:
SelfInstall Leap Micro test should boot iso instead of raw images.
More info in https://bugzilla.opensuse.org/show_bug.cgi?id=1197877


mloviska:
@jlausuch @nilxam

jlausuch (approved):
LGTM!

jlausuch:
@andrii-suse @Martchus could you please merge?

andrii-suse:
Hm, it looks `Default-SelfInstall` is not synced anymore and has no `isos post` as well. It is certainly possible that such combination is not supported by generation, I will have a look

andrii-suse (changes requested):
Ah it looks the xml has `raw.xz` for Default-SelfInstall , but .lst file has `iso` : `Default-SelfInstall-Build16.23.install.iso` . Which one is correct?

mloviska:
> Ah it looks the xml has `raw.xz` for Default-SelfInstall , but .lst file has `iso` : `Default-SelfInstall-Build16.23.install.iso` . Which one is correct?

Made a mistake there, I should update only `SelfInstall` flavors. Thanks!

andrii-suse (approved):

PR #166, opened 2022-05-10
Align SeflInstall flavors with SLEM

mloviska:
Matching flavors are required in order to reuse easily schedules between sle-micro and leap-micro.


mloviska:
`openSUSE-Leap-Micro.x86_64-5.2.0-Default-RT-Build16.23.raw.xz` was a copy&paste issue, this image should not be downloaded anymore.

andrii-suse (approved):

PR #167, opened 2022-05-25
Remove EoL Leap WSL schedule

mloviska:
~~WSL Leap image are now compress in [`xz` format](https://build.opensuse.org/package/binaries/openSUSE:Leap:15.4:WSL/kiwi-images-wsl/images)~~

~~Leap 15.4 is using python3-kiwi ==> 9.24.17-150100.3.50.1 which is
missing a fix for this issue. As of now, it is not sure when this will
be resolved.~~

The initial problem has been resolved, so we just need to remove the EoL schedule


andrii-suse:
Since you xml is deleted for WSL 15.2 - to make the integration test happy you shall clean WSL 15.2 from few locations in t/docker as well, otherwise LGTM.

PR #170, opened 2022-06-16
Start scheduling ALP product

mloviska:


xml/obs/devel:LEO.xml
```diff
@@ -0,0 +1,18 @@
+<openQA
+    project_pattern="devel:LEO"
+    dist_path="images"
+    distri="alp"
+    archs="x86_64">
```
jlausuch, line 5: I know that @ggardet is planning to work on arm images, maybe we could also include `aarch64` here already? Will the fail if it can't find those images?

mloviska: hm... good point, I will try to make it work for existing images and then add `aarch64` so everything is prepared ahead

ggardet: Yes, please include aarch64 asap please :)

jlausuch (approved):
Neat! The isos post look alright. This looks promising :)

xml/obs/devel:LEO.xml
```diff
@@ -0,0 +1,26 @@
+<openQA
+    project_pattern="devel:LEO"
+	dist_path="obspublish-other::openqa/devel:LEO/images/x86_64"
```
andrii-suse, line 3: I think it is safe to replace `x86_64` with `*` here if we target several architectures, e.g. openSUSE:Leap:15.2:MicroOS.xml does the same.

andrii-suse (approved):

jlausuch (approved):
LGTM

jlausuch:
@andrii-suse can we merge this?

andrii-suse:
Please note that for automatic triggering the OBS project must be added to a list somewhere in openSUSE-release-tools , let me know if I need to find where.

andrii-suse:
automatic sync is done for projects which prefixes are listed in gocd/monitors.gocd.yaml :  https://github.com/openSUSE/openSUSE-release-tools/blob/67d6535836432c5aea512c95fc2de79598a28adb/gocd/monitors.gocd.yaml#L124

mloviska:
> @mloviska _fcrozat_ told me he will enable `aarch64` for `images` as soon as the current build will be done.
> 
> But, shouldn't we have a single FTP tree for all ALP architectures, as it is done for Leap?

Good point, but I have no idea as I haven't set up the project :D

PR #171, opened 2022-06-21
Disable ARM images in ALP syncing

mloviska:
ARM images are not ready yet, therefore `SelfInstall` images are not being pulled even for `x86_64`.


jlausuch (approved):
LGTM. As agreed with @fcrozat, we disable aach64 until it's built, otherwise, it fails the pipeline.

jlausuch:
> aarch64 images and FTP tree are now built. PR #175

https://github.com/os-autoinst/openqa-trigger-from-obs/pull/176

PR #172, opened 2022-06-29
Revert "Add JeOS and MicroOS images to openSUSE:Factory:Staging"

DimStar77:
Reverts os-autoinst/openqa-trigger-from-obs#157

That blocks all stagings completely from being synced now. BAck to the drawing board


DimStar77:
FYI @Vogtinator @andrii-suse

PR #173, opened 2022-07-06
Fix regression caused by #170

jlausuch:
Instead of forcing the `/` we provide it in the xml as a suffix.


jlausuch:
@mloviska @andrii-suse please review. 
See https://github.com/os-autoinst/openqa-trigger-from-obs/pull/170/files#r913499597 for more info.

andrii-suse (approved):
Yeah sorry - I had to check it more carefully. `read_files.sh` is the weakest spot that is not covered by tests.

jlausuch:
@andrii-suse sorry, I forgot to update the xml .. https://github.com/os-autoinst/openqa-trigger-from-obs/pull/174

PR #174, opened 2022-07-06
Fix regression caused by #170  (fix devel:LEO xml)

jlausuch:
Instead of forcing the `/` we provide it in the xml as a suffix.

### DIFF
--- a/xml/obs/devel:LEO.xml
+++ b/xml/obs/devel:LEO.xml
@@ -8,7 +8,7 @@
 		<flavor name="SelfInstall" folder="ALP:SelfInstall">
 			<hdd filemask="ALP.*-SelfInstall-Build.*install.iso$"/>
 			<repos>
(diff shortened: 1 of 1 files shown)


andrii-suse:
Hm, I am not sure if it is really needed. After `make test_regen_all` I see in `read_files.sh`: `rsync -4 --list-only $rsync_pwd_option obspublish-other::openqa/devel:LEO/images/*//*ALP-ftp-POOL*/ ` - it is ending on `/` already

jlausuch:
> Hm, I am not sure if it is really needed. After `make test_regen_all` I see in `read_files.sh`: `rsync -4 --list-only $rsync_pwd_option obspublish-other::openqa/devel:LEO/images/*//*ALP-ftp-POOL*/ ` - it is ending on `/` already

Right, I tested locally too... not sure why it was introduced in https://github.com/os-autoinst/openqa-trigger-from-obs/pull/170/files then

PR #175, opened 2022-07-08
Enable aarch64 images in ALP syncing

ggardet:
Enable aarch64 images in ALP syncing since the images and FTP tree are now built for aarch64 as well.


ggardet:
Closed in favor to #176

PR #177, opened 2022-08-08
Leap micro53

mloviska:


andrii-suse (approved):
One minor thing: If you keep old .lst files for 5.2 (or maybe use separate commit for them) - the diff would be smaller. Otherwise lgtm

mloviska:
> One minor thing: If you keep old .lst files for 5.2 (or maybe use separate commit for them) - the diff would be smaller. Otherwise lgtm

You are fast! :) Alright, next time I will keep this in my mind. Thanks a lot Andrii

PR #179, opened 2022-08-18
Add openSUSE:Leap:15.5:ARM:Images

Vogtinator:
A copy of 15.4 again. The files still contain 15.4 in their names, the version
wasn't bumped yet.


andrii-suse (approved):

andrii-suse:
AFIR version 15.4 vs 15.5 inside .lst files doesn't matter, so once the file names contain 15.5 - no action is needed and sync should just work.

PR #189, opened 2022-12-06
Retry failing docker tests

kalikiana:
See: https://progress.opensuse.org/issues/120468

### DIFF
--- a/t/run_docker.sh
+++ b/t/run_docker.sh
@@ -5,7 +5,9 @@ dockerfail=0
 cd docker
 for t in *.sh; do
     [ -x "$t" ] || continue
-    ./$t
+    for i in $(seq 1 3); do
+        ./$t && s=0 && break || s=$? && sleep 10;
+    done
     if [ $? -eq 0 ] ; then
         echo "P:$t"
     else


okurz (approved):
Could be fine and is simple enough. However I would like to use 
https://build.opensuse.org/package/show/openSUSE:Factory/retry but it's not in Leap. I found that our CI image is still Leap 15.3 so I created
https://github.com/os-autoinst/openqa-trigger-from-obs/pull/190 to at least bring it up to Leap 15.4 as well as https://github.com/os-autoinst/openqa-trigger-from-obs/pull/191 to immediately go to Tumbleweed. That would also allow us to install "retry" directly.

Martchus (approved):

t/run_docker.sh
```diff
@@ -5,7 +5,9 @@ dockerfail=0
 cd docker
 for t in *.sh; do
     [ -x "$t" ] || continue
-    ./$t
+    for i in $(seq 1 3); do
+        ./$t && s=0 && break || s=$? && sleep 10;
+    done
     if [ $? -eq 0 ] ; then
```
perlpunk, line 11: Hm, I think `$?` will always be 0 here, and the actual return code we want is stored in `$s`, no?

perlpunk: pened #193

PR #191, opened 2022-12-07
CI: Update to Tumbleweed to not need to care about Leap version updates

okurz:


andrii-suse:
I don't think that testing in only TW is a good idea, especially that things like apparmor are involved. Ideally we must test on the same system which is in production, TW should be optional addition.

PR #192, opened 2022-12-07
Add aarch64 to ALP

ggardet:
Requires https://github.com/os-autoinst/opensuse-jobgroups/pull/233


ggardet:
I think the CI fails because there is a mismatch in build numbers in OBS.

Martchus:
This change also affects many x86-related lines. I'm not very familiar with how all of this works but are you sure?

ggardet:
> This change also affects many x86-related lines. I'm not very familiar with how all of this works but are you sure?

It looks like it removes duplicates only.

andrii-suse:
> It looks like it removes duplicates only.

Exactly. Presence of .license folders in files_repo.lst started causing those duplicates. Adding `$` at https://github.com/os-autoinst/openqa-trigger-from-obs/pull/192/files#diff-4d9910c8d9edac07b3908da7c83a0f329a41581b7282d170d73559ac3b1e7d90R162 now removes them.

PR #198, opened 2023-02-17
Schedule Leap micro 5.4

mloviska:
- ticket: https://progress.opensuse.org/issues/124134


andrii-suse:
Strictly speaking it is not necessary to have these commits if xml is the same and the commands are the same and only version changed.
We just need to create new folder on o3 and generate scripts like `make openSUSE:Leap:Micro:5.4:Images:ToTest` and maybe remove old 5.2 and 5.3 folders.

So while it is not required to push these, it may be easier to track eventual changes in future, so LGTM.

mloviska:
> Strictly speaking it is not necessary to have these commits if xml is the same and the commands are the same and only version changed. We just need to create new folder on o3 and generate scripts like `make openSUSE:Leap:Micro:5.4:Images:ToTest` and maybe remove old 5.2 and 5.3 folders.
> 
> So while it is not required to push these, they may help to track eventual changes in future, so LGTM.

Yup, that is true, my intention was to keep it clean. :) 
Thanks

lkocman:
👍

PR #201, opened 2023-03-08
Add new ALP Micro images

jlausuch:


jlausuch:
@andrii-suse @foursixnine @mloviska please have a look

mloviska (approved):

foursixnine (approved):

jlausuch:
As the OBS failures are taking some time to be fixed... Let's just merge this and see what happens with the image triggering in O3 and do the adjustments later on.

andrii-suse:
I've merged, but then realized that in order to enable automatic sync:
the folder and `project_pattern` in xml must have name of obs project, so `SUSE:ALP:Products:Micro:0.1`  instead of `SUSE:ALP:Micro`

jlausuch:
> I've merged, but then realized that in order to enable automatic sync: the folder and `project_pattern` in xml must have name of obs project, so `SUSE:ALP:Products:Micro:0.1` instead of `SUSE:ALP:Micro`

even we use specific `dist_path` ?

jlausuch:
> yeah dist_path is to where take files. It can be even url iirc and by default it is the name of folder, but may be some custom path as well. But in order to receive correct rabbitmq notifications we need to know explicitly name of OBS project to subscribe to. (maybe we can try to extract it from dist_path, but not sure if it is good idea and it is better to have clear understanding when scripts should trigger looking only at the folder name).

ok, done here: https://github.com/os-autoinst/openqa-trigger-from-obs/pull/202
thanks!

PR #206, opened 2023-03-22
Add ALP Bedrock images

jlausuch:


jlausuch:
> just retriggered the CI

seems that `download.opensuse.org` is a bit slow today...

grisu48 (approved):
The XML and the `*.lst` files looks sane.

PR #210, opened 2023-03-26
Leap Micro has different build numbers for isos and images

jlausuch:
The xml was merged into single one:
https://github.com/os-autoinst/openqa-trigger-from-obs/pull/208 The project contains the string ToTest, which is normally used in Factory, but ToTest is also used for Leap Micro,  so the script does not allow having different build numbers in the files.


jlausuch:
This is the output before applying this change
```
geekotest@ariel:/opt/openqa-trigger-from-obs> bash -x script/rsync.sh openSUSE:Leap:Micro:5.4:ToTest
+ set -e
+ environ=openSUSE:Leap:Micro:5.4:ToTest
+++ dirname script/rsync.sh
++ cd script
++ pwd
+ DIR=/opt/openqa-trigger-from-obs/script
+ cd /opt/openqa-trigger-from-obs/script/..
+ '[' -d openSUSE:Leap:Micro:5.4:ToTest ']'
+ '[' -e openSUSE:Leap:Micro:5.4:ToTest/rsync.lock ']'
+ trap 'rm -f openSUSE:Leap:Micro:5.4:ToTest/rsync.lock; exit' INT TERM EXIT
+ echo 13759
+ '[' '!' -f rsync.secret ']'
++ cat rsync.secret
+ export RSYNC_PASSWORD=tZNMx.gSU38NE3EV
+ RSYNC_PASSWORD=tZNMx.gSU38NE3EV
+ failures_count=0
+ for subfolder in $environ/*/
+ '[' -d 'openSUSE:Leap:Micro:5.4:ToTest/*/' ']'
+ subfolder=openSUSE:Leap:Micro:5.4:ToTest
+ set +e
+ set -e
+ '[' -e openSUSE:Leap:Micro:5.4:ToTest/read_files.sh ']'
+ '[' '!' -e openSUSE:Leap:Micro:5.4:ToTest/print_rsync_iso.sh ']'
+ '[' -e openSUSE:Leap:Micro:5.4:ToTest/print_openqa.sh ']'
+ bash -e openSUSE:Leap:Micro:5.4:ToTest/read_files.sh
+ '[' -e openSUSE:Leap:Micro:5.4:ToTest/.run_last ']'
+ [[ openSUSE:Leap:Micro:5.4:ToTest == *ToTest* ]]
+ [[ openSUSE:Leap:Micro:5.4:ToTest != *Factory* ]]
+ [[ openSUSE:Leap:Micro:5.4:ToTest != *MicroOS* ]]
++ grep -h -o -E 'Build[0-9](\.|[0-9]+)*[0-9]+' openSUSE:Leap:Micro:5.4:ToTest/files_iso.lst 'openSUSE:Leap:Micro:5.4:ToTest/files_repo_000product:Leap-Micro-ftp-ftp*.lst' 'openSUSE:Leap:Micro:5.4:ToTest/Media1_000product:Leap-Micro-ftp-ftp*.lst'
+ builds='Build3.23
Build3.23
Build3.23
Build3.23
Build13.2
Build13.2
Build14.2
Build14.2'
+ '[' -n 'Build3.23
Build3.23
Build3.23
Build3.23
Build13.2
Build13.2
Build14.2
Build14.2' ']'
++ echo 'Build3.23
Build3.23
Build3.23
Build3.23
Build13.2
Build13.2
Build14.2
Build14.2'
++ sort
++ uniq
++ wc -l
+ '[' 3 -gt 1 ']'
+ echo 'Conflicting builds found {Build3.23
Build3.23
Build3.23
Build3.23
Build13.2
Build13.2
Build14.2
Build14.2}, exiting because of conflict in {openSUSE:Leap:Micro:5.4:ToTest}'
Conflicting builds found {Build3.23
Build3.23
Build3.23
Build3.23
Build13.2
Build13.2
Build14.2
Build14.2}, exiting because of conflict in {openSUSE:Leap:Micro:5.4:ToTest}
+ exit 0
+ res=0
+ '[' 0 -eq 0 ']'
+ '[' openSUSE:Leap:Micro:5.4:ToTest '!=' openSUSE:Leap:Micro:5.4:ToTest ']'
+ break
+ exit
+ rm -f openSUSE:Leap:Micro:5.4:ToTest/rsync.lock
+ exit
```

jlausuch:
After trying this, I get new errors: 
https://openqa.opensuse.org/admin/obs_rsync/openSUSE:Leap:Micro:5.4:ToTest/runs/.run_230326_213641_13.2/download/generate_openqa.err

andrii-suse:
So CI is correct that scripts generate inconsistent commands: in the sync that has been triggered aarch64 folder is created:
                Leap-Micro-5.4-aarch64-Build13.2, 
but in openqa command it is referenced as
REPO_0=Leap-Micro-5.4-aarch64-Build3.23

It is possible that the test doesn't need REPO variable and succeeds, but the folder will be deleted by openQA assets on next run (it deletes every repo, which is not referenced by any test).

The scripts are not expected to run with different build id for different architectures, at least there was no attempt to run them before. I will have a look what can be done.

jlausuch:
> So CI is correct that scripts generate inconsistent commands: in the sync that has been triggered aarch64 folder is created: Leap-Micro-5.4-aarch64-Build13.2, but in openqa command it is referenced as REPO_0=Leap-Micro-5.4-aarch64-Build3.23
> 
> It is possible that the test doesn't need REPO variable and succeeds, but the folder will be deleted by the automatic job (of openQA assets cleanup) on next run. Because it deletes every repo, which is not referenced by any test.
> 
> The scripts are not expected to run with different build id for different architectures, at least there was no attempt to run them before. I will have a look what can be done.

The build number is the same for different architectures in the same flavor (DVD-x86_64 and DVD-aarch64). 
The difference is in DVD vs Images. This is the same case for SLE Micro, where we combine both in the same place, but we don't use REPO_X there as we have proxy scc. 


I see that 
`rsync --timeout=3600 -rtlp4 --delete --specials --link-dest /var/lib/openqa/factory/repo/Leap-Micro-5.4-x86_64.license-CURRENT/ /var/lib/openqa/factory/repo/Leap-Micro-5.4-x86_64.license-CURRENT/ /var/lib/openqa/factory/repo/Leap-Micro-5.4-x86_64.license-Build13.2`
is right, but the actual problem I see is that the REPO_LEAP variable points to the wrong build number:
`REPO_LEAP_MICRO=Leap-Micro-5.4-aarch64-Build13.2 `

andrii-suse:
Ah, I see now. The DVD files were just deleted from files_iso.lst in #208 , so CI allowed the run:  https://github.com/os-autoinst/openqa-trigger-from-obs/pull/208/files#diff-81261a81443178a1b917136e161d9c0c79834451235b4186d0f91b0d01cfce68L1-L2

I will check if it is simple to provide a patch.

jlausuch:
> Ah, I see now. The DVD files were just deleted from files_iso.lst in #208 , so CI allowed the run: https://github.com/os-autoinst/openqa-trigger-from-obs/pull/208/files#diff-81261a81443178a1b917136e161d9c0c79834451235b4186d0f91b0d01cfce68L1-L2
> 
> I will check if it is simple to provide a patch.

Thanks!

jlausuch:
Thank you for the help, Andrii.

Vogtinator:
Yes, this approach will not work with ttm.

Vogtinator:
There are three options:

a) Revert this change
b) Merge the `openSUSE:Leap:Micro:5.4` and `:Images` projects on OBS, then use setrelease to have a single release.
c) Add some new complexity into ttm to explicitly look for this version or change the code to behave like rabbit-openqa.py which looks for `ISO=` matching instead of `BUILD=` matching.

AFAIK, the purpose of the separate `:Images` project in OBS is to allow images to get rebuild separately also after final release. That makes merging the project both on OBS and in the syncing invalid and thus a) is the only correct option.

jlausuch:
Ok. I will then revert all the changes and move to the previous state. BUT the implication is that the images will use the repo they defined by default, not any repo from the ISO project. 

I would like that for the next version we go for option b) as we do for SLE Micro (images and isos are in the same IBS project, and they get different build numbers.)

nilxam:
maybe we need to revert https://github.com/os-autoinst/openqa-trigger-from-obs/pull/208 also?

jlausuch:
> maybe we need to revert #208 also?

yes, https://github.com/os-autoinst/openqa-trigger-from-obs/pull/213
but first I Wanted to revert the latest change. 
and ofc https://github.com/os-autoinst/opensuse-jobgroups/pull/301

nilxam:
> Ok. I will then revert all the changes and move to the previous state. BUT the implication is that the images will use the repo they defined by default, not any repo from the ISO project.
> 
> I would like that for the next version we go for option b) as we do for SLE Micro (images and isos are in the same IBS project, and they get different build numbers.)

we can revisit this topic in the next Leap Micro, since Leap Micro 5.4 entered Beta stage(close to RC almost) I would appareciate that  we can keep the previous  working approach :)

PR #200, opened 2023-03-03
schedule openSUSE Leap 15.4 Continous Rebuild

DeepthiYV:
Related Ticket : https://progress.opensuse.org/issues/120243

PR #216, opened 2023-06-13
Revert "schedule openSUSE Leap 15.4 Continous Rebuild"

Vogtinator:
Reverts os-autoinst/openqa-trigger-from-obs#200

This uses the wrong project, it MUST use `openSUSE:Leap:15.4:Images:ToTest` instead.


Martchus (approved):

foursixnine:
@DeepthiYV is taking a look at this; I doubt we need to revert, we could however add a CI check to always look for projects that conform the same build structure, so the assets should always be expected in the `*:ToTest` subproject...

@DimStar77, @fcrozat wdyt? (for both, IBS and OBS, giving some time for IBS projects to transition)

Vogtinator:
> we could however add a CI check to always look for projects that conform the same build structure, so the assets should always be expected in the *:ToTest subproject...

What should work is: If it starts with `openSUSE:`, it has to either contain `:Staging:`, `:Rings:` or end in `:ToTest`.

However, I think this error is rather rare and shouldn't really happen so maintaining the check is more work then it would save.

foursixnine:
> I fail to see any relevance with SLE or ALP here..

We use a similar project on our internal GitLab, so we make it a QA Requirement that projects will have to have a `ToTest` subproject, the same way it's already being done with ALP (See here https://gitlab.suse.de/openqa/openqa-trigger-from-ibs-plugin/-/commit/6b4acc8280e26311265f05564ec5959aed563290#77f0ac33670ed997edf266204620373f90ab2834_3_3)

The intention being here: is that for us (at SUSE and openSUSE), there's a "friendly suggested way" how to set up projects for testing, giving at least some level of consistency.

fcrozat:
Currently, no project is setup the same way, either on openSUSE nor on SUSE (ALP is different from SLE and SLE products are not always using the same concepts as Factory in terms of setup).

Vogtinator:
> Currently, no project is setup the same way, either on openSUSE nor on SUSE

?

Literally all openSUSE projects are set up the same way and have been for ages

foursixnine:
> Currently, no project is setup the same way, either on openSUSE nor on SUSE (ALP is different from SLE and SLE products are not always using the same concepts as Factory in terms of setup).

Yes, that's what I'm proposing. 

> However, I think this error is rather rare and shouldn't really happen so maintaining the check is more work then it would save.

I've seen it a few times already, at least enough to bother me :)

Vogtinator:
> > However, I think this error is rather rare and shouldn't really happen so maintaining the check is more work then it would save.
> 
> I've seen it a few times already, at least enough to bother me :)

Oh ok. This is the first time I've seen it.

Vogtinator:
Included in #218

PR #230, opened 2023-08-24
Add openSUSE:Leap:15.6:Images

Vogtinator:
Like 15.5:Images. Some images didn't get the version bump yet and the Vagrant Box for VirtualBox is missing.


foursixnine (approved):

andrii-suse:
just to note: if nothing changes in xml / scripts for 15.6 vs 15.5 - there is no big need to add new files to t/ folder. 
(But it will not hurt as well and will help once some new changes in generation are done)

Vogtinator:
> just to note: if nothing changes in xml / scripts for 15.6 vs 15.5 - there is no big need to add new files to t/ folder. (But it will not hurt as well and will help once some new changes in generation are done)

IMO the rule of thumb to just always add a test is not bad. At some point, the older versions will be removed, along with their tests. If the new versions don't have tests on their own, they would either end up untested or the removal requires adding tests...

PR #234, opened 2023-09-21
Fix syncing .license repos

andrii-suse:


andrii-suse:
@Vogtinator do you know if the openQA test use hardcoded REPO_2 and REPO_3 ? (If yes - then this needs some more work).

Vogtinator:
> @Vogtinator do you know if the openQA test use hardcoded REPO_2 and REPO_3 ? (If yes - then this needs some more work).

I don't think any `REPO` is hardcoded anywhere in medium types, testsuites  or job groups.

Vogtinator:
Does anything actually use this `.license` "repo"? I don't think this is even possible, so maybe it should just not be synced at all.

andrii-suse:
It is hard to confirm. IIRC I was expelicitly asked to add it to some of internal projects. My understanding it is accessible by appending '.license' to corresponding value of REPO_... variable(s). I will merge it now at least to stop spreading bloat on disk and network in o3 sync.

Vogtinator:
> It is hard to confirm. IIRC I was expelicitly asked to add it to some of internal projects. My understanding it is accessible by appending '.license' to corresponding value of REPO_... variable(s). I will merge it now at least to stop spreading bloat on disk and network in o3 sync.

Who asked for this? I don't see any test code in the os-autoinst-distri-opensuse repo that does this. Maybe some SUT software references it somehow, but in that case I'd also like to know how.

PR #238, opened 2023-12-21
Do not consider failure to sync fatal

kalikiana:
We still log the error but we assume it is okay to fail.

See: https://progress.opensuse.org/issues/112871


script/rsync.sh
```diff
@@ -36,7 +36,7 @@ set +e
     [ -e "$subfolder/print_openqa.sh" ] || { >&2 echo "No file found: {$subfolder/print_openqa.sh}"; exit 1; }
 
     # nowhere to log yet as we haven't created $logfolder
-    bash -e "$subfolder/read_files.sh" ||  { >&2 echo "read_files.sh failed for $subfolder in enviroment $environ"; exit 1; }
+    bash -e "$subfolder/read_files.sh" || { >&2 echo "read_files.sh failed for $subfolder in enviroment $environ"; exit 0; }
```
okurz, line 39: If we do this then we should be more explicit about the impact, i.e. at least rewording the text string to make it clear that the error is basically ignored. But maybe that would be better to decide within read_files.sh and not here in a general way.

perlpunk (commented):
Hm, but this would ignore any error in read_files.sh?
I don't have a good idea how to find out the cause, but ignoring all errors might hide real problems in the future

perlpunk:
So read_files.sh is doing
```
rsync -4 --list-only $rsync_pwd_option dist.suse.de::repos/SUSE:/SLE-15-SP6:/GA:/Staging:/A/images/repo/ | grep -P '...' | awk '...' | grep -v IGNOREPATTERN | grep -E 'Module-Basesystem|...' | grep -E '...'  >> .../files_repo.lst
```
and if one of the grep fails, the script fails.
Maybe there is a way to make only the grep commands non-fatal.

perlpunk:
This should work:
```
rsync ... (grep ... | grep ... >>logfile || true)
```
I tested it with a mocked rsync.
So if rsync has a problem, the script would still fail.

okurz:
How would this look in reality? I would like to see two cases, one with error and one without. And what about the explicit log message we planned?

perlpunk:
This is a line from a genererated read_files.sh:
```
rsync -4 --list-only $rsync_pwd_option dist.suse.de::repos/SUSE:/SLE-15-SP6:/Update:/BCI/images//*repo/SLE-BCI-**
 | grep -P 'Media1(.license)?$'
 | awk '{ $1=$2=$3=$4=""; print substr($0,5); } '
 | grep -v IGNOREPATTERN
 | grep -E '' 
| grep -E 'aarch64|ppc64le|s390x|x86_64'  >> /opt/openqa-trigger-from-ibs/SUSE:SLE-15-SP6:Update:BCI/files_repo_SLE-BCI-*.lst
```
And I would guess that some of the grep commands is failing, but we probably want to ignore the return value of grep here.

script/scriptgen.py
```diff
@@ -674,7 +674,7 @@ def gen_read_files(self, f):
         if self.assets and self.assets_flavor:
             for k, v in self.asset_folders.items():
                 self.p(
-                    """rsync -4 --list-only $rsync_pwd_option PRODUCTPATH/{}/*{} | awk '{{ $1=$2=$3=$4=""; print substr($0,5); }}' >> __envsub/files_asset.lst""".format(
+                    """rsync -4 --list-only $rsync_pwd_option PRODUCTPATH/{}/*{} | awk '{{ $1=$2=$3=$4=""; print substr($0,5); }}' >> __envsub/files_asset.lst || true""".format(
```
kalikiana, line 677: Discussed in jitsi. This may be the wrong part of the generator, see https://github.com/os-autoinst/openqa-trigger-from-obs/pull/238#issuecomment-1886916332 - I will look into that after coming up with a fake rsync that produces successful or error output, and see how grep and awk respond to that.

kalikiana: So it seems the `read_files_repo` value in `cfg.py:23` is the basis for that rsync call and the `p` function in `scriptgen.py` actually replaces the target file that is being piped into i.e. `__Envsub/files_repo.lst`

Martchus (approved):
I'm guess I still don't get the full picture of how this repo works but the tests look like we've discussed (and you found an even better way to mock the commands).

PR #239, opened 2024-01-23
Add Slowroll image testing

bmwiedemann:


okurz (approved):

xml/obs/openSUSE:Slowroll:Staging-images.xml
```diff
@@ -0,0 +1,10 @@
+<openQA
+    project_pattern="openSUSE:Slowroll:Staging"
+    dist_path="images"
```
b10n1k, line 3: i think this has to be `images/x86_64`. bouble check because i am not sure

xml/obs/openSUSE:Slowroll:Staging-images.xml
```diff
@@ -0,0 +1,10 @@
...
+    distri="opensuse"
+    archs="x86_64">
+    <flavor name="JeOS-for-kvm-and-xen" folder="*/kiwi-templates-Minimal:kvm-and-xen">
+        <!-- Due to there being multiple archs, it omits "head -n1". So don't match .sha256(.asc). -->
+        <hdd filemask=".*kvm-and-xen.*\.qcow2$"/>
+    </flavor>
```
b10n1k, line 9:
```suggestion
    <flavor name="Slowroll" distri="opensuse" iso="1" folder="*product*"/>
```
I would expect something like that. Why flavor is `JeOS-for-kvm-and-xen`?

okurz: Basically the flavor can be selected arbitrary just that according openQA scenario definitions or job templates need to exist. For example for Leap 15.5 Images https://openqa.opensuse.org/admin/job_templates/103 shows sections with key "products" like "opensuse-Leap15.5-JeOS-for-kvm-and-xen-x86_64"

xml/obs/openSUSE:Slowroll:Staging-images.xml
asdil12: updated.

Martchus (approved):

PR #241, opened 2024-04-24
Store uncompressed vmdk in HDD folder

mloviska:
As we had no uncompressed vmdk images before, these images would land in other directory instead of hdd.


jlausuch:
LGTM, thanks!

PR #243, opened 2024-05-07
Allow repos without Media1 prefix as required for Marble

andrii-suse:


jlausuch (approved):
LGTM. Thanks a lot for this

PR #244, opened 2024-05-15
Update for new Slowroll location

bmwiedemann:


bmwiedemann:
Somehow this still does not trigger runs, even though there is
http://download.opensuse.org/repositories/openSUSE:/Slowroll:/Build:/1/images/?P=*kvm-and-xen*qcow2

PR #249, opened 2024-06-13
Example hook script

Martchus:
See particular commit messages and https://progress.opensuse.org/issues/161879

### DIFF
--- a/script/rsync.sh
+++ b/script/rsync.sh
@@ -79,6 +79,12 @@ set +e
     [ ! -L "$subfolder/.run_last" ] || rm "$subfolder/.run_last"
     ln -s -T "$(pwd)/$logdir" $subfolder/.run_last
 
+    # run additional hook script from current subfolder
+    [[ -x "$subfolder/hook_script" ]] && "$subfolder/hook_script" 2>$logdir/hook_script.err > $logdir/hook_script.out
+
+    # run additional hook script from script folder
+    [[ -x "$DIR/generic_hook" ]] && "$DIR/generic_hook" "$environ" 2>$logdir/generic_hook.err > $logdir/generic_hook.out
+
     [ ! -e "$subfolder/print_openqa.sh" ] || bash -e "$subfolder/print_openqa.sh" 2>$logdir/generate_openqa.err > $logdir/openqa.cmd
 
     for f in {rsync_iso.cmd,rsync_repo.cmd,openqa.cmd}; do
--- a/script/generic_hook
+++ b/script/generic_hook
@@ -1,0 +1,9 @@
+# !/bin/bash
+
+echo "Hello work hook script run for OBS project '$1'."
+
+# you could query the OBS API from here
+
+# you could compute a YAML document containing relevant test scenario (see https://open.qa/docs/#scenarios_yaml)
+
+# you could output a file called `openqa.cmd` with the openqa-cli commands to be run (not using `print_openqa.sh` for that anymore)


script/rsync.sh
```diff
@@ -79,6 +79,12 @@ set +e
     [ ! -L "$subfolder/.run_last" ] || rm "$subfolder/.run_last"
     ln -s -T "$(pwd)/$logdir" $subfolder/.run_last
 
+    # run additional hook script from current subfolder
+    [[ -x "$subfolder/hook_script" ]] && "$subfolder/hook_script" 2>$logdir/hook_script.err > $logdir/hook_script.out
+
+    # run additional hook script from script folder
+    [[ -x "$DIR/generic_hook" ]] && "$DIR/generic_hook" "$environ" 2>$logdir/generic_hook.err > $logdir/generic_hook.out
+
```
okurz, line 87: this looks like it would be what we need but to have that mergeable the example script above shouldn't be commited directly. Instead you could put it into a documentation in the README of this project.

script/rsync.sh
```diff
@@ -79,6 +79,12 @@ set +e
     [ ! -L "$subfolder/.run_last" ] || rm "$subfolder/.run_last"
     ln -s -T "$(pwd)/$logdir" $subfolder/.run_last
 
+    # run additional hook script from current subfolder
+    [[ -x "$subfolder/hook_script" ]] && "$subfolder/hook_script" 2>$logdir/hook_script.err > $logdir/hook_script.out
```
okurz, line 83: Can the hook_script have any effect on the command line that is generated and executed afterwards? How about to allow to export variables in the hook script which are read if defined by the later scripts to be able to customize the generated isos sync command(s)?

script/rsync.sh
Martchus: Yes, this is just a draft to show what it would look like for real.

kalikiana: I would consider 106 lines way too much for an inline code snippet.

script/rsync.sh
Martchus: I think the established pattern here is that scripts generate certain output and those files and then be consumed later on. So I would just have the hook script generated e.g. a file with scenario definitions and then maybe also directly the `openqa.cmd` file with the `openqa-cli` commands that will be invoked later by `rsync.sh`.

Not sure how it would work with environment variables. We cannot just source the hook script because it might also be e.g. a Python script.

foursixnine: @Martchus, think that a JSON data structure would be the lingua franca for hooks to exchange information. Following up on that, the solution may go in a different direction.

@Martchus, I'm happy to chat and brainstorm in the office or during a call

script/rsync.sh
okurz: @kalikiana not script/rsync.sh which is not a hook script but script/generic_hook which is 9 lines

script/rsync.sh
Martchus: > think that a JSON data structure would be the lingua franca for hooks to exchange information. Following up on that, the solution may go in a different direction.

Note that the consumer of the hook script output is also just a script (`print_openqa.sh` if present; otherwise `rsync.sh` will just invoke `openqa.cmd` if present).  So probably it makes most sense to simply generate the required `openqa-cli` commands and write them to `openqa.cmd` and skip having `print_openqa.sh`.

It is really *not* like there's some subsequent logic in openQA that would be able to consume a JSON file. There's just the generation of `openqa-cli` commands within this trigger repo itself and so far this is just a generated shell script.

> I'm happy to chat and brainstorm in the office or during a call

We'll probably have a chance to talk in the office as of the 24th.

PR #255, opened 2024-07-25
Add openqa_cli property to xml

andrii-suse:
The idea is that some projects can put custom wrapper instead of openqa-cli calls.
That wrapper can manage generated parameters as needed, make additional calls, etc.

### DIFF
--- a/script/cfg.py
+++ b/script/cfg.py
@@ -219,7 +219,7 @@ def openqa_call_start_meta_variables(meta_variables):
 def pre_openqa_call_start(repos):
     return ''
 
-openqa_call_start = lambda distri, version, archs, staging, news, news_archs, flavor_distri, meta_variables, assets_flavor, repo0folder: '''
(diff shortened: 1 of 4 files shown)


andrii-suse:
Do not merge until the plugin is adjusted as well

Martchus:
This might be a good alternative to https://github.com/os-autoinst/openQA/pull/5757 (which is rather messy anyway, at least in its current form). So I guess the related ticket is https://progress.opensuse.org/issues/164388.

foursixnine:
> Do not merge until the plugin is adjusted as well

@andrii-suse what else is needed here? or you mean that some changes in another repo are needed?

andrii-suse:
> @andrii-suse what else is needed here? or you mean that some changes in another repo are needed?

I've checked - the other repo should be fine, so this is good to go, feel free to merge

okurz:
Let's try then :)

PR #260, opened 2024-08-26
Reapply Leap 16.0

lkocman:
* Use unique package name agama-installer-Leap for Leap, so we do not conflict with Devel and Tumbleweed 
* Was reverted in #259


lkocman:
Related product change in https://github.com/os-autoinst/opensuse-jobgroups/pull/512 and openQA /admin/products /admin/groups

PR #263, opened 2024-08-30
Leap16 add repos

andrii-suse:


lkocman (approved):
All looks good! I would not even think of touching cfg.py :-)

PR #268, opened 2024-09-11
Add systemd timer and service for updates

kalikiana:
See: https://progress.opensuse.org/issues/166433


Martchus (approved):

Martchus:
Maybe we should apply @baierjan's suggestions about the targets in the other PR here as well.

PR #269, opened 2024-09-12
Adjust install targets of systemd service and timer

kalikiana:
See: https://progress.opensuse.org/issues/166433

### DIFF
--- a/systemd/openqa-trigger-from-obs-update.service
+++ b/systemd/openqa-trigger-from-obs-update.service
@@ -6,6 +6,3 @@ After=network-online.target
 Type=oneshot
 ExecStart=/usr/bin/git -C /opt/openqa-trigger-from-obs pull --quiet --rebase origin master
 User=geekotest
-
-[Install]
-WantedBy=default.target
--- a/systemd/openqa-trigger-from-obs-update.timer
+++ b/systemd/openqa-trigger-from-obs-update.timer
@@ -6,4 +6,4 @@ OnCalendar=*:0/3
 Unit=openqa-trigger-from-obs-update.service
 
 [Install]
-WantedBy=default.target
+WantedBy=timers.target


Martchus (approved):

okurz:
@kalikiana @Martchus is this still needed? I don't see a good justification, just "adjust". So … is this a fix, correction, improvement?

systemd/openqa-trigger-from-obs-update.timer
```diff
@@ -6,4 +6,4 @@ OnCalendar=*:0/3
 Unit=openqa-trigger-from-obs-update.service
 
 [Install]
-WantedBy=default.target
+WantedBy=timers.target
```
Martchus, line 9: This makes sense according to https://man.archlinux.org/man/systemd.special.7. We also use this in other timer-triggered units.

systemd/openqa-trigger-from-obs-update.service
```diff
@@ -6,6 +6,3 @@ After=network-online.target
 Type=oneshot
 ExecStart=/usr/bin/git -C /opt/openqa-trigger-from-obs pull --quiet --rebase origin master
 User=geekotest
-
-[Install]
-WantedBy=default.target
```
Martchus, line 11: This means one can no longer enable this unit normally, e.g. to run the triggering just once after boot. I guess this makes sense for this timer-triggered unit. We also don't have this for other timer-triggered units.

Martchus (approved):
I think both changes make sense and that's also what we have in similar units. (I've just  checked *all* timers and their service files in the openQA repo.)

kalikiana:
> @kalikiana @Martchus is this still needed? I don't see a good justification, just "adjust". So … is this a fix, correction, improvement?

This refers to a comment on #268 after merging. Somehow the follow-up got lost, so thank you for catching that!

PR #272, opened 2024-09-17
Add agama integration test and tweak the apparmor profile

andrii-suse:


andrii-suse:
My goal was to add a test for apparmor, which was a good way to test the profiles. But apparently apparmor doesn't do anything inside containers anymore. So I guess I will merge this when next time I look at agama or similar

PR #274, opened 2024-10-21
Dont add SLE_ prefix to repo in openQA if it starts with SL

andrii-suse:
This will break the plugin MR, so they will need to regenerate the projects

### DIFF
--- a/script/scriptgen.py
+++ b/script/scriptgen.py
@@ -1089,7 +1089,13 @@ done < <(sort __envsub/files_asset.lst)""",
         if self.repos or self.repolink:
             if self.ag.brand == "ibs":
                 self.p(" i=9", f)  # temporary to simplify diff with old rsync scripts, may be removed later
-            self.p(cfg.openqa_call_repot(self.build_id_from_iso, self.repos), f, "REPOTYPE", "''", "REPOPREFIX", "SLE_")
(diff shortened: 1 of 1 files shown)


andrii-suse:
see https://gitlab.suse.de/openqa/openqa-trigger-from-ibs-plugin/-/merge_requests/206/diffs

script/scriptgen.py
```diff
@@ -1089,7 +1089,12 @@ def gen_print_openqa(self, f):
         if self.repos or self.repolink:
             if self.ag.brand == "ibs":
                 self.p(" i=9", f)  # temporary to simplify diff with old rsync scripts, may be removed later
-            self.p(cfg.openqa_call_repot(self.build_id_from_iso, self.repos), f, "REPOTYPE", "''", "REPOPREFIX", "SLE_")
+            # some trickery for REPO_SLE_ vs REPO_SL_ variables in ibs
+            sl = "SLE_"
+            for r in self.repos:
+                if r.tag.startswith("SL"):
+                    sl = ""
```
Martchus, line 1096: Maybe a break would make sense here as well, or any like in this example: https://stackoverflow.com/questions/2012611/any-function-in-python-with-a-callback

andrii-suse: sure, it was a test for whether anybody tries to follow this madness :)

jlausuch (approved):
LGTM

PR #276, opened 2024-11-05
Micro 6.1: Fix format for encrypted images raw -> raw.xz

lkocman:
This (and fact that RT images were not building) was blocking sync from OBS to openQA. This should fix it 

First imported build: https://openqa.opensuse.org/group_overview/102


lkocman:
Many thanks to @andrii-suse for final debugging!

PR #278, opened 2024-11-21
[WIP] Support extract_as_repo for SLM pxe tar file

alice-suse:


alice-suse:
@jlausuch @andrii-suse Welcome to review!

foursixnine (approved):
@rfan1 you were also looking at providing this too right?

rfan1:
> @rfan1 you were also looking at providing this too right?

Yes, but I was looking for the extracted iso images for ppc64le arch only to start the selfinstall tests.

andrii-suse (approved):
hm, do you have a confirmation from openQA devs that putting .tar to `factory/iso` is a good idea? If you are sure about that, then looks good to me

script/cfg.py
```diff
@@ -72,9 +72,11 @@ def rsync_commands(checksum):
         asset_folder=other
         [[ ! $dest =~ \.iso$  ]] || asset_folder=iso
         [[ ! $dest =~ \.spdx\.json$  ]] || asset_folder=iso
+        [[ ! $dest =~ \.tar$  ]] || asset_folder=iso
```
foursixnine, line 75: @andrii-suse looking at your comment should this be:

```suggestion
        [[ ! $dest =~ \.tar$  ]] || asset_folder=repo
```

alice-suse: That's the original default logic, putting it to `other`, but it brings problem elsewhere.  Originally the code only supports extract iso into repo, so for anywhere related with such action, it assumes the image is put in `iso/...`. If changing it, it means a lot of work to dig into other code parts and fix problems... Given that there is already `[[ ! $dest =~ \.spdx\.json$  ]] || asset_folder=iso`, I think it will be easier to just putting tar into iso folder. 

I also thought about another way to implement this, like completely support any other format that needs to be  extracted as repo, just like iso, but it will require too much  effort from top(parsing settings) to down(generating scripts), which is not proper for me who is just freshman to this tool. So I chose the lower effort solution, like pretending to be "iso"...

If you guys think this is acceptable, I am happy to see it merged. Otherwise, I am fine to close the PR and maybe let other expert guys on the tool to implement the function in a more comprehensive way.

foursixnine: >  I think it will be easier to just putting tar into iso folder.

Makes sense

> which is not proper for me who is just freshman to this tool.

by now I consider you an expert xD - I barely understand how this all works 📦 

@andrii-suse do you know if its possible to delete an asset or could you provide proposal on to of Alice's change, so that the .tar file is removed after extraction?

alice-suse: ISO files are kept in iso folder after extraction. I think the same is better to be applied to tar files -- just to be synced in behaviors. HDYT?

andrii-suse: > If you guys think this is acceptable, I am happy to see it merged. 

Yeah, the questions are whether it will work as planned and whether it can bring any problems.

> Otherwise, I am fine to close the PR

That is nobody's goal here. You have already made decent work here, let's focus on make it through if that is really needed.
I will investigate a bit more and get back to this.

foursixnine: > ISO files are kept in iso folder after extraction. I think the same is better to be applied to tar files -- just to be synced in behaviors. HDYT?

I'm fine :)

alice-suse: > > If you guys think this is acceptable, I am happy to see it merged.
> 
> Yeah, the questions are whether it will work as planned and whether it can bring any problems.
> 

Good point! 

I believe you can get some useful information from this MR : 
- export SLM pxe tarball and publish as repo: https://gitlab.suse.de/openqa/openqa-trigger-from-ibs-plugin/-/merge_requests/213

You will see how it works for SLM tar image. And after full script re-generation, it does not affect other project directories' *.before. However, for the areas which are not touched in xml/*bs/*, we have to rely on you experts to tell :).

Martchus: I was asked about how well openQA handles `.tar` assets as `ISO`:

* It looks like openQA uses the asset extension in some places to determine the asset type. That will break in cases the asset type is important it might not behave as required.
* The cache service might also not consider the asset judging by its code which filters assets by their extensions (`my $assets = $tree->map('to_string')->grep(qr/\.(?:img|qcow2|iso|vhd|vhdx)$/);`).
* Specifying a tar archive via the `ISO` variable is also problematic because os-autoinst backends will handle it like a cd/dvd image. That will probably not work and might lead to errors.

I'd say you can try it nevertheless but don't be surprised if it doesn't work on some level.

PR #283, opened 2025-02-05
Cleanup old Leap EOL projects

dzedro:
https://progress.opensuse.org/issues/175704


xml/obs/openSUSE:Leap:15.xml
```diff
@@ -1,12 +0,0 @@
-<openQA
```
grisu48, line 1: This file looks generic for all Leap 15.* versions to me. Sure this also needs to be removed?

grisu48: Same comment for the next files as well, I think the `Leap:15:` files need to stay?

dzedro: I noticed that too

dzedro: `project_pattern="openSUSE:Leap:(?P&lt;version&gt;15.[3-9]):ToTest"`
restored the `:15:` files

grisu48 (approved):

xml/obs/openSUSE:Leap:15.xml
grisu48: Thank you!

PR #288, opened 2025-03-07
Add global attribute media1 to xml

andrii-suse:


andrii-suse:
See also https://gitlab.suse.de/openqa/openqa-trigger-from-ibs-plugin/-/merge_requests/238

jlausuch (approved):

PR #289, opened 2025-03-19
Avoid warning about escape sequence in python 3.12

perlpunk:
https://docs.python.org/dev/whatsnew/3.12.html#other-language-changes

    /opt/openqa-trigger-from-ibs/script/scriptgen.py:489: SyntaxWarning: invalid escape sequence '\.'
        """build=$(grep -o -E '(Build|Snapshot)[^-]*' __envsub/files_iso.lst | grep -o -E '[0-9]\.?[0-9]+(\.[0-9]+)*' | head -n 1)""",

### DIFF
--- a/script/scriptgen.py
+++ b/script/scriptgen.py
@@ -486,7 +486,7 @@ class ActionBatch:
     def gen_repo(self, repodir, gen, f):
         if "$build" in gen:
             self.p(
-                """build=$(grep -o -E '(Build|Snapshot)[^-]*' __envsub/files_iso.lst | grep -o -E '[0-9]\.?[0-9]+(\.[0-9]+)*' | head -n 1)""",
+                r"""build=$(grep -o -E '(Build|Snapshot)[^-]*' __envsub/files_iso.lst | grep -o -E '[0-9]\.?[0-9]+(\.[0-9]+)*' | head -n 1)""",
                 f,
             )
         body = gen


okurz (approved):

script/scriptgen.py
```diff
@@ -486,7 +486,7 @@ def gen_if_not_customized(self, folder, fname):
     def gen_repo(self, repodir, gen, f):
         if "$build" in gen:
             self.p(
-                """build=$(grep -o -E '(Build|Snapshot)[^-]*' __envsub/files_iso.lst | grep -o -E '[0-9]\.?[0-9]+(\.[0-9]+)*' | head -n 1)""",
+                """build=$(grep -o -E '(Build|Snapshot)[^-]*' __envsub/files_iso.lst | grep -o -E '[0-9]\\.?[0-9]+(\\.[0-9]+)*' | head -n 1)""",
```
b10n1k, line 489: why not use the suggested raw string? just add `r` in the beginning of the string! wouldnt that work? I think it would simplify the expression. Adding slashes it is harder to read sometimes. wdyt?

b10n1k (changes requested):

script/scriptgen.py
perlpunk: sure, i don't mind much. It doesn't seem to break anything else, so done

PR #291, opened 2025-04-30
Restrict Leap DVD to Leap 15.5+

grisu48:
There is no point in syncing Leap 15.3 and Leap 15.4 anymore.


Vogtinator:
That's a noop - those projects don't trigger events anyway. It won't do anything with DVD-Updates as that's a different flavor managed by a different bot

grisu48:
Closing as it would not make a difference anyways.

PR #293, opened 2025-05-12
Update regex to include build number in leap 16

foursixnine:
See discussion in 
- https://suse.slack.com/archives/C02CANHLANP/p1747036806061369


foursixnine:
This is not needed anymore after #295

PR #299, opened 2025-06-26
Rename iso list for Leap 16.0

lkocman:
We're now using Leap_16 flavor instead of Leap, and the version is back to agama version.


grisu48 (commented):
I'm missing the changes to the corresponding xml file in here, is that correct? :thinking:

PR #300, opened 2025-07-01
Re-org iso list for Leap 16.0

nilxam:
intend to supersedes https://github.com/os-autoinst/openqa-trigger-from-obs/pull/299 also


lkocman (approved):
looks good. Thank you @nilxam

PR #305, opened 2025-07-23
Add openqa/public needed by openqa-cli to apparmor

kalikiana:
The file openqa-cli.yaml needs to be readable.

See: https://github.com/os-autoinst/openQA/pull/6316


profiles/apparmor.d/opt.openqa-trigger-from-obs.script.rsync.sh
```diff
@@ -76,6 +76,7 @@
     /opt/openqa-trigger-from-obs/*:*/.run*/openqa*.log w,
     /usr/bin/openqa-cli rix,
     /usr/share/openqa/lib/** r,
+    /usr/share/openqa/public/** r,
```
perlpunk, line 79: I would suggest to just allow the whole directory:
```suggestion
    /usr/share/openqa/** r,
```

PR #307, opened 2025-07-28
Switch Factory x86 from FTP tree to product composer

foursixnine:
See https://progress.opensuse.org/issues/186426

Looking at the `0NonFree` and the `0openSUSE` directories its clear where the location is

```
# rsync --timeout=3600 -tlp4 --specials --list-only obspublish::openqa/openSUSE:Factory:ToTest/*images/local/*0openSUSE/
drwxr-xr-x          4,096 2025/07/28 13:21:34 .
-rw-r--r--     10,226,379 2025/07/27 17:30:36 .meta.success
-rw-r--r--              1 2025/07/28 13:21:27 .setrelease
-rw-r--r--        301,411 2025/07/27 18:23:39 _buildenv
-rw-r--r--          1,031 2025/07/27 18:23:39 _statistics
-rw-r--r--     26,536,131 2025/07/27 18:23:39 logfile
-rw-r--r--     10,593,758 2025/07/27 18:23:44 openSUSE-20250727-x86_64-Debug.report
-rw-r--r--      6,707,558 2025/07/27 18:23:40 openSUSE-20250727-x86_64-Source.report
-rw-r--r--     29,510,574 2025/07/27 18:23:39 openSUSE-20250727-x86_64.cdx.json
-rw-r--r--     23,340,246 2025/07/27 18:23:43 openSUSE-20250727-x86_64.report
-rw-r--r--     50,752,991 2025/07/27 18:23:39 openSUSE-20250727-x86_64.spdx.json
drwxr-xr-x             46 2025/07/28 13:21:28 openSUSE-20250727-x86_64-Debug
drwxr-xr-x             59 2025/07/28 13:21:29 openSUSE-20250727-x86_64-Source
drwxr-xr-x          4,096 2025/07/28 13:21:29 openSUSE-20250727-x86_64.license
drwxr-xr-x          4,096 2025/07/28 13:21:28 openSUSE-20250727-x86_64
```

PR #308, opened 2025-07-30
Fix read_files.sh for Factory/base

andrii-suse:
followup from #307

### DIFF
--- a/script/scriptgen.py
+++ b/script/scriptgen.py
@@ -659,11 +659,19 @@ class ActionBatch:
                 suffix = ""
                 if repodir.attrib.get("suffix"):
                     suffix = repodir.attrib["suffix"]
+                repopath = ""
+                if "/" in self.folder:
+                    repopath = self.ag.productpath + "/" + self.folder + "/*" + repodir.attrib["folder"] + "*" + suffix
+                else:
+                    repopath = (
+                        self.ag.productrepopath() + "/" + self.folder + "/*" + repodir.attrib["folder"] + "*" + suffix
+                    )
+
                 args = (
                     cfg.read_files_repo,
                     f,
                     txt,
-                    self.ag.productpath + "/" + self.folder + "/*" + repodir.attrib["folder"] + "*" + suffix,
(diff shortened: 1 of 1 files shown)


script/scriptgen.py
```diff
@@ -701,21 +709,6 @@ def gen_read_files(self, f):
...
-                            f,
-                            "PRODUCTREPOPATH",
-                            self.ag.productpath + "/" + self.folder + "/*" + repodir.attrib["folder"] + wild,
-                            "Media1.lst",
-                            "",
-                            "destlst",
-                            "Media1_{}.lst".format(os.path.basename(repodir.attrib["folder"]).lstrip("*")),
-                            "media.1/media",
-                            "media.1/products",
-                        )
```
foursixnine, line 718: Won't this break projects that haven't switched yet?

andrii-suse: No it is only health check for Factory and LEO when repo path didn't have snapshot in it

foursixnine (approved):

script/scriptgen.py
```diff
@@ -659,11 +659,19 @@ def gen_read_files(self, f):
                 suffix = ""
                 if repodir.attrib.get("suffix"):
                     suffix = repodir.attrib["suffix"]
+                repopath = ""
+                if "/" in self.folder:
+                    repopath = self.ag.productpath + "/" + self.folder + "/*" + repodir.attrib["folder"] + "*" + suffix
+                else:
+                    repopath = (
+                        self.ag.productrepopath() + "/" + self.folder + "/*" + repodir.attrib["folder"] + "*" + suffix
+                    )
```
b10n1k, line 668:
```suggestion
    if "/" in self.folder:
        prd_path = self.ag.productpath
    else:
       prd_path = self.ag.productrepopath()
       glob_pattern = f"*{repodir.attrib['folder']}*{suffix}"
       repopath = (os.path.join(prd_path, self.folder, glob_pattern))
```
I think this reduce the duplication

PR #313, opened 2025-07-31
Fix read_files.sh for Factory/live

andrii-suse:


script/scriptgen.py
```diff
@@ -361,9 +361,9 @@ def doFlavor(self, node):
                 self.isos.append(iso)
                 if node.attrib.get("ln_iso_to_repo", ""):
                     self.ln_iso_to_repo[iso] = node.attrib["ln_iso_to_repo"]
-
-        if node.attrib.get("name", "") and node.attrib.get("folder", ""):
-            self.iso_folder[node.attrib["name"]] = node.attrib["folder"]
+        elif node.attrib.get("name", "") and node.attrib.get("folder", ""):
+            for iso in node.attrib["name"].split("|"):
+                self.iso_folder[node.attrib["name"]] = node.attrib["folder"]
```
Martchus, line 366: That the loop variable isn't used in the loop body seems wrong. Should this be like `self.iso_folder[iso] = node.attrib["folder"]` above? (The *intention* of this code is really hard to make sense of.)

PR #321, opened 2025-09-17
Prevent openQA from cleaning up `CURRENT` repositories

Martchus:
* Sync `CURRENT` repos to `fixed` subdirectory
* See https://progress.opensuse.org/issues/188667


andrii-suse (approved):

Martchus:
Regarding AppArmor: It looks like we already use the double-asterisks like `/var/lib/openqa/share/factory/repo/** rwl` in all occurrences of `repo` in the profile in this repo and in `/etc/apparmor.d/opt.openqa-trigger-from-obs.script.rsync.sh` on o3. So I *think* we already cover granting permissions in subdirectories like `repo/fixed`.

okurz (approved):

PR #322, opened 2025-10-06
Include livepatches variables in openqa.cmd

d3flex:
Automate the livepatch isos posts. Due to the logic of the script generation this commit contains an injection of a for loop for all the livepatches which will be found in the spdx.json, after the sync in the server.

- the loop always run even if no livepatches found. However it will only append to the isos posts the variable in the case that actually livepatch versions inclue in the array. The approach is just a bit dirty, as any other approach would require further and bigger changes.
- The injection is in the `rsync_iso` after all the necessary variables are set and before the beginning writing to the print_openqa.sh. At that point the `spdx.json` should be available and it needs only to grep for the _kernel_livepatch_ in the server. Before that point, the scriptgen.py would have to make a request to each repo in the dist.s.d and it is likely it would be less flexible as the logic of the _isos post_ is mainly in the generated `print_openqa.sh`.

depends-on: https://gitlab.suse.de/openqa/openqa-trigger-from-ibs-plugin/-/merge_requests/263
issue: https://progress.opensuse.org/issues/190134

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ jobs:
       - checkout
       - run:
           command: |
-            zypper -n in --no-recommends diffutils make python3 tar git-core
+            zypper -n in --no-recommends diffutils make python3 tar git-core python3-requests python3-beautifulsoup4
       - checkout
       - run: git clean -df
       - run: make test_regen_all
--- a/script/cfg.py
+++ b/script/cfg.py
@@ -1,4 +1,8 @@
 import os
+import re
+import requests
+from bs4 import BeautifulSoup
+from datetime import datetime
 
 header = '''# GENERATED FILE - DO NOT EDIT
 set -e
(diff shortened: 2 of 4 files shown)


mdoucha:
> > Can you update the ticket reference?
> > The use of `NONE` seems also needlessly complicated. Why not just use an empty array?
> 
> this will make the for runs once, netherless if the `grep` return something or not. That was to avoid bigger changes with if else and duplication as there are different pieces in different fuctions which construct the final script.

The base `isos post` command has to run once every time, even when there are livepatches. Also, you schedule all livepatch `isos post` with the same default `BUILD` value. You need to add the kernel version and flavor to `BUILD`.

> > This approach will schedule tests for livepatch packages which haven't been updated in the product increment.
> 
> In the ticket, you said that `This is the product increment repository with new packages to test + everything already released (in SCC): https://download.suse.de/ibs/SUSE:/ALP:/Products:/Marble:/6.0:/ToTest/product/repo/SL-Micro-6.0-x86_64/`. The obs-sync already syncs this repo with openQA https://openqa.suse.de/admin/obs_rsync/SUSE:ALP:Products:Marble:6.0:ToTest/runs/.run_last/download/rsync_iso.cmd
> 
> What am I missing?

You're missing that the repo contains everything that was already released. We don't want to test everything, we want to test only the packages that are new since the last product increment.

PR #325, opened 2025-11-25
Add agama based net-install for TW on s390

foursixnine:
- **Sync agama media for s390**
- **Cleanup openSUSE:Factory:zSystems:ToTest**
- **Add files for s390/base**
- **Add before files for net-agama**


DimStar77:
agama-installer:openSUSE fails to build on s390x (yay) - the PXE flavor is deprecated and should no longer be used

PR #358, opened 2026-04-29
Ignore armv6hl

foursixnine:
This should help with syncing ARM builds and should help with [poo#196478](https://progress.opensuse.org/issues/196478)


okurz:
@r-richardson

PR #363, opened 2026-05-19
Enable XML-driven exclusions for rsync calls

r-richardson:
Introduces an optional `exclude` attribute to XML repository definitions, allowing selective filtering of the main rsync payload (which previously could not be filtered).

- Updates generator templates to pass XML excludes to rsync commands.
- Applies `exclude="*armv6l* *armv6hl*"` to openSUSE:Factory:ARM to reduce payload size, preventing timeouts.
- Adds missing `armv6hl` to the global debug/source exclusion list.

Related Progress Ticket: https://progress.opensuse.org/issues/196478


script/cfg.py
```diff
@@ -270,19 +273,19 @@ def rsync_repodir1_dest_media0(dest, debug, source, folder):
...
 '''
     res = res + '''
             continue
         }
 '''
 
     res = res + '''
-        echo rsync --timeout=3600 -rtlp4 --delete --specials PRODUCTREPOPATH/''' + folder + '''/$src/  /var/lib/openqa/factory/repo/fixed/$dest-CURRENT/
-        echo rsync --timeout=3600 -rtlp4 --delete --specials --link-dest /var/lib/openqa/factory/repo/fixed/$dest-CURRENT/ /var/lib/openqa/factory/repo/fixed/$dest-CURRENT/ /var/lib/openqa/factory/repo/$dest-$buildid
+        echo "rsync --timeout=3600 -rtlp4 --delete --specials ''' + xtra + ''' PRODUCTREPOPATH/''' + folder + '''/$src/  /var/lib/openqa/factory/repo/fixed/$dest-CURRENT/"
```
Wabri, line 287: question(if-minor): I'm not a fan of `'` spam, what if we do something like this:

```
res = res + f'''
        echo "rsync --timeout=3600 -rtlp4 --delete --specials {xtra} PRODUCTREPOPATH/{folder}/$src/  /var/lib/openqa/factory/repo/fixed/$dest-CURRENT/"
'''
```

Definitely more eyes pleasing. Since there are a lot of this echo line, I don't know if it's worth the cost of replacing it.

script/cfg.py
```diff
@@ -236,9 +236,12 @@ def rsync_repomultiarch(destpath, debug, source):
         [[ ! $src =~ .*\\.license ]] || dest=$dest.license'''
 
 
-def rsync_repodir1_dest_media0(dest, debug, source, folder):
+def rsync_repodir1_dest_media0(dest, debug, source, folder, exclude=""):
     repo = os.path.basename(folder).lstrip("*")
     xtra=""
+    if exclude:
+        for ex in exclude.split(" "):
```
Wabri: question(blocking): why not a use `split()` like in the other part of the pr?

Wabri (changes requested):

script/cfg.py
r-richardson: I agree that this looks a bit cleaner, but if we use f-strings here i think then we also have to escape the bash syntax curly braces by doubling them (since python f-strings use {} to inject variables). For example, earlier in the same function we would have to change `${arch//i686/i586}` to `${{arch//i686/i586}}`, and `${archs[@]}` to `${{archs[@]}}`, which imo is also not pretty, and would increase the likelihood of typos during future edits.

script/cfg.py
r-richardson: right, its inconsistent and its probably better to split on any kind of whitespace instead of just space. changed :+1:

asmorodskyi (approved):

okurz (approved):

Wabri (approved):

PR #364, opened 2026-05-19
fix(style): resolve ruff E701 and E731 violations, add type checking

d3flex:
Replace black with ruff and ty for Python style enforcement. The `requires-python = ">=3.10"` constraint in pyproject.toml applies only to `uv run` dev tooling, which runs on Tumbleweed in the CircleCI style job. Runtime scripts in script/ remain compatible with Python 3.6 on Leap 15.5 mostly, however the image was updated from Leap 15.5 to Leap 16.0 (which runs also python 3.13).

Changes:
- E701 (multiple statements on one line): split compound conditions and function definitions across multiple lines in script/cfg.py and script/scriptgen.py
- E731 (lambda assignment): convert rsync_iso lambda to a proper function definition in script/cfg.py
- Replace triple-quoted strings with double-quote delimiters throughout script/cfg.py for consistency
- Small refactoring in script/scriptgen.py: replace `is not` None chains with `any()` for readability when checking multiple optional parameters

Issue: https://progress.opensuse.org/issues/191575

### DIFF
--- a/script/abs.py
+++ b/script/abs.py
@@ -1,7 +1,6 @@
-import scriptgen
 import cfg
 
-cfg.header = '''# GENERATED FILE - DO NOT EDIT
+cfg.header = """# GENERATED FILE - DO NOT EDIT
 set -e
 # abstract test config
-'''
+"""
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2.1
 jobs:
   unit:
     docker:
-      - image: registry.opensuse.org/opensuse/leap:15.5
+      - image: registry.opensuse.org/opensuse/leap:16.0
     steps:
       - checkout
       - run:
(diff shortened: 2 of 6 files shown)


script/abs.py
```diff
@@ -1,7 +1,6 @@
-import scriptgen
 import cfg
 
-cfg.header = '''# GENERATED FILE - DO NOT EDIT
+cfg.header = """# GENERATED FILE - DO NOT EDIT
```
Wabri, line 3: nitpick(blocking): There is an uppercase `DO NOT EDIT` here 😆

Wabri (changes requested):

script/abs.py
Wabri: Ups, it's the string that is going to be generated... Sorry! 😆

Wabri (approved):

okurz (approved):

PR #366, opened 2026-05-22
fix(style): resolve ruff E701 and E731 violations, add type checking

d3flex:
Replace black with ruff and ty for Python style enforcement. The `requires-python = ">=3.10"` constraint in pyproject.toml applies only to `uv run` dev tooling, which runs on Tumbleweed in the CircleCI style job. Runtime scripts in script/ remain compatible with Python 3.6 on Leap 15.5 mostly, however the image was updated from Leap 15.5 to Leap 16.0 (which runs also python 3.13).

Changes:
- E701 (multiple statements on one line): split compound conditions and function definitions across multiple lines in script/cfg.py and script/scriptgen.py
- E731 (lambda assignment): convert rsync_iso lambda to a proper function definition in script/cfg.py
- Replace triple-quoted strings with double-quote delimiters throughout script/cfg.py for consistency
- Small refactoring in script/scriptgen.py: replace `is not` None chains with `any()` for readability when checking multiple optional parameters

Issue: https://progress.opensuse.org/issues/191575


okurz (approved):
Approved. jlausuch should approve after their manual verification

Makefile
```diff
@@ -38,4 +38,6 @@ test_update_before_files: test_regen_all
 	(cd t && bash test_before_after_diff.sh --update-before *bs/*)
 
 test_python_style:
-	black --check --fast --diff script/scriptgen.py 
+	uv run ruff check
+	uv run ruff format --check
+	uv run ty check
```
Martchus: Please don't require the use of `uv` in Makefile targets. You can make the `uv run` part configurable via an env variable.

Martchus (changes requested):

okurz (approved):

Martchus (approved):

PR #372, opened 2026-06-16
Revert Leap 16.0 to syncing ftp tree only

foursixnine:
Follows #369 and #370


nilxam:
leave the approval to @nilxam

PR #385, opened 2026-08-05
fix(script/cfg.py): Avoid stale/broken symlinks when unpacking the iso

foursixnine:
Prevent a silent failure and nested directory link creation when the
destination target is a physical directory or a resolved symbolic link
to a directory.

Force `ln` to fail if the target is somehow converted into a directory
(e.g manual intevention)

IBS will need: https://gitlab.suse.de/openqa/openqa-trigger-from-ibs-plugin/-/merge_requests/295

### DIFF
--- a/script/cfg.py
+++ b/script/cfg.py
@@ -105,10 +105,12 @@ for flavor in {FLAVORLIST,}; do
         + (repo0folder if repo0folder else "")
         + """
         iso_repo="/var/lib/openqa/factory/repo/${repo0folder}"
+        iso_repo_current_alias="/var/lib/openqa/factory/repo/fixed/VERSIONVALUE-${flavor}-${arch}-CURRENT"
         [ -z "FLAVORASREPOORS" ] || [ $( echo "$flavor" | grep -E -c "^(FLAVORASREPOORS)$" ) -eq 0 ] || echo "[ -d "${iso_repo}" ] || {
     mkdir $iso_repo
     bsdtar xf /var/lib/openqa/factory/iso/$dest -C $iso_repo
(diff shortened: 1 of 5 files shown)


script/cfg.py
```diff
@@ -108,7 +108,7 @@ def rsync_iso(
         [ -z "FLAVORASREPOORS" ] || [ $( echo "$flavor" | grep -E -c "^(FLAVORASREPOORS)$" ) -eq 0 ] || echo "[ -d "${iso_repo}" ] || {
     mkdir $iso_repo
     bsdtar xf /var/lib/openqa/factory/iso/$dest -C $iso_repo
-    ln -sf "$iso_repo" "/var/lib/openqa/factory/repo/fixed/VERSIONVALUE-${flavor}-${arch}-CURRENT"
+    ln -fsT "$iso_repo" "/var/lib/openqa/factory/repo/fixed/VERSIONVALUE-${flavor}-${arch}-CURRENT"
```
alice-suse: Thanks for the help.
If I understand correctly, this will force it to fail when "/var/lib/openqa/factory/repo/fixed/VERSIONVALUE-${flavor}-${arch}-CURRENT" exists already. It is indeed better than creating nested link under it. But the link can not be ensured successful. How about deleting the directory before linking?

foursixnine: My thought here was to follow how the project behaves when somebody modifies a script and removes the header: Do nothing.

If somebody going to the server and creating a directory exactly there, is because there's a reason for it, so the rsync script will fail and nothing will be triggered.

But we can say that manual intervention on generated stuff is not supported :P. The code now deletes the directory/link before creating a new one

alice-suse (approved):
LGTM, thank you.

PR #387, opened 2026-08-12
fix(script/cfg.py): Only treat known disk image suffixes as HDD assets

KuxaBeast:
## Problem

The rsync destination folder is chosen in `rsync_isos()` by matching the file suffix:

```bash
[[ ! $dest =~ \.(qcow2|raw|vhd|vmdk|vhdx|xz)$ ]] || asset_folder=hdd
```

The trailing `xz` is a catch-all, so **any** compressed file ending in `.xz` is rsynced into `$factory/hdd/`.

The openQA variable name is derived independently, by `openqa_call_start_ex()`, which maps only an explicit list of disk image suffixes to `HDD_1`:

```
\.(hdd|qcow2|raw|raw\.xz|raw\.gz|vhdx\.xz|vmdk|vmdk\.xz)$
```

and falls back to `ASSET_1` for everything else.

The two classifiers therefore disagree for any `.xz` file that is not a disk image. Such a file is copied to `$factory/hdd/`, while openQA is told `ASSET_1=<file>` and resolves it under `$factory/other/`. The asset is never found and the jobs fail. The `.sha256` sibling already lands in `$factory/other/` correctly, which makes the mismatch easy to overlook.

## Fix

Split the single rule in two and restrict the compressed variant to the same disk image suffixes `openqa_call_start_ex()` knows about, so both classifiers agree:

```bash
[[ ! $dest =~ \.(qcow2|raw|vhd|vmdk|vhdx)$ ]] || asset_folder=hdd
[[ ! $dest =~ \.(qcow2|raw|vhd|vmdk|vhdx)\.(xz|gz)$ ]] || asset_folder=hdd
```

## Impact

`.raw.xz` is the only compressed suffix currently produced by any flavor (8 occurrences across `t/obs/*/*/files_iso.lst`). It keeps its `$factory/hdd/` destination, so **no test fixture changes** and `make test` stays at 162 passed / 0 failures.

The bug surfaced while preparing the switch of the openSUSE Factory `WSL` flavor from the `.appx` image to the `.tar.xz` one produced by `kiwi-images-tbz` ([poo#203721](https://progress.opensuse.org/issues/203721)). That change is intentionally kept out of this PR and will be submitted separately; it exercises this code path and is the first consumer of a non-disk-image `.xz` asset.

### DIFF
--- a/script/cfg.py
+++ b/script/cfg.py
@@ -95,8 +95,7 @@ for flavor in {FLAVORLIST,}; do
         [[ ! $dest =~ \.iso$  ]] || asset_folder=iso
         [[ ! $dest =~ \.spdx\.json$  ]] || asset_folder=iso
         [[ ! $dest =~ \.tar$  ]] || asset_folder=iso
-        [[ ! $dest =~ \.(qcow2|raw|vhd|vmdk|vhdx)$ ]] || asset_folder=hdd
-        [[ ! $dest =~ \.(qcow2|raw|vhd|vmdk|vhdx)\.(xz|gz)$ ]] || asset_folder=hdd
+        [[ ! $dest =~ \.(qcow2|raw|vhd|vmdk|vhdx)(\.(xz|gz))?$ ]] || asset_folder=hdd
         """
         + rsync_commands(checksum)
         + """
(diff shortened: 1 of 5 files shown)


ricardobranco777 (approved):

Martchus (dismissed):

foursixnine (commented):
Can you add tests? I guess by touching the xml and the files_iso.lst would do

- xml/abs/TestIsoHdd.xml
- t/abs/TestIsoHdd-1/files_iso.lst

foursixnine (changes requested):

mpagot (approved):

ricardobranco777 (approved):

foursixnine (approved):

PR #389, opened 2026-08-14
feat(wsl): Test Factory WSL against the tarball image instead of the appx

KuxaBeast:
Point the Factory `WSL` flavor at the tarball image instead of the appx.

`openSUSE:Factory/kiwi-images-tbz` already builds a `*.tar.xz` for every
snapshot on both architectures, copies it into `:ToTest` next to the appx and
publishes it to the appliances directory on download.opensuse.org, so this is
only a matter of picking the other package up.

### Impact

The flavor name stays `WSL`, so the existing scenarios pick the tarball up as
`ASSET_1` without any job group change. `BUILD` is still scraped from the
`Snapshot<date>` segment of the file name and is unaffected. The tarball keeps
landing in `factory/other`, since it is not a disk image - that part depends on
#387.

The unrelated `.tar.xz` files published in the same appliances directory (the
LXC and dnf images) cannot match, because the rsync path is scoped to the OBS
package folder.

Related:
- https://progress.opensuse.org/issues/203721
- https://github.com/os-autoinst/opensuse-jobgroups/pull/904
- https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/26403


foursixnine (changes requested):
@KuxaBeast Sorry to be pedantic here, could you ammend the commit? the commit message looks longer than it should, and has extra context that isn't really necessary for somebody working with this particular project.

mpagot (approved):

Martchus (approved):

foursixnine (approved):

